PR #46, opened 2015-02-05
Fixes #45 by giving civix the non-dereferenced path to the extension.

### LINKED ISSUE #45: Civix throws Exception if extension is symlinked into the site
Exception: Failed to locate CMS. Please call civix from somewhere under the CMS root.

```
me@mine:/usr/local/lib/civicrm-buildkit/build/drupal-clean$ ls -hal sites/default/ext/
total 16K
drwxrwxr-x+ 2 me me 4.0K Feb  5 14:39 .
drwxr-xr-x  5 me me 4.0K Jan 30 15:55 ..
lrwxrwxrwx  1 me me   64 Feb  5 14:35 com.ginkgostreet.groovy -> /home/me/projects/com.ginkgostreet.groovy/
```

Civix *is* being called from under the root; it just doesn't know it :-)

GinkgoFJG:
I didn't dig through the rest of the codebase to see which other `getcwd()`'s it might be appropriate to replace with `exec("pwd")`'s, but this seems to work.

### DIFF
--- a/src/CRM/ClientBundle/ClientFactory.php
+++ b/src/CRM/ClientBundle/ClientFactory.php
@@ -16,7 +16,7 @@ class ClientFactory {
   public function get() {
     $origDir = getcwd();
 
-    list ($cmsRoot, $civicrmConfigPhp) = $this->findCivicrmConfigPhp(getcwd());
+    list ($cmsRoot, $civicrmConfigPhp) = $this->findCivicrmConfigPhp(exec("pwd"));
     if (!is_dir($cmsRoot)) {
       throw new \Exception('Failed to locate CMS. Please call civix from somewhere under the CMS root.');
     }


totten:
Clever. :)

veda-consulting:
This amendment breaks windows machines! If we revert the amend everything works as expected. Any other potential workarounds to the issue you were trying to fix?

GinkgoFJG:
What do you mean it breaks Windows machines? What kind of error or behavior are you getting? Here's [the issue I was trying to fix](https://github.com/totten/civix/issues/45).

What does `exec("pwd")` return on a Windows machine? If it returns something consistent (like FALSE or NULL), maybe we can write a wrapper function which falls back to `getcwd()`.

totten:
FWIW, I have a long-term WIP PR for civicrm-core which borrowed the exec("pwd") technique... and had the same thought. It's untested, though:

https://github.com/civicrm/civicrm-core/pull/5114/files#diff-713ced7ff8be6f7c6411a623b34804eaR310

```php
 if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN') {
   return getcwd();
 }
 else {
   exec('pwd', $output);
   return trim(implode("\n", $output));
 }
```

GinkgoFJG:
Dang, I didn't know about the PHP_OS constant. That sounds better than what I suggested.

veda-consulting:
Sounds like a plan (windows machines return an error basically as the pwd command doesn't exist).
I'll get our guys to try this patch and see if it works for them.

antrik:
I'd say it would be better to check whether the pwd command actually fails, rather than doing OS-specific case handling?

(Also, I'm pretty sure Windows does have some 'pwd' equivalent...)

PR #48, opened 2015-02-25
Additional features for the civix test command.

anemirovsky:
I added a couple of additional options that can be passed to the civix test
command to make it easier to run and debug tests.

The first thing I did was make the Test Class argument optional because...

I needed a way to run all of the test classes for my extension at once and
couldn't find a good way to do this as currently implemented. So, I added a
--configuration option that will use PHPUnit's --configuration option to run
the tests as defined in a phpunit.xml file in the root of the extension.

I also added a --debug option as I've found the output with that option
incredibly helpful for tracking long test runs.

Finally, I added --colors option as I like the color output for an immediate
notification of pass or failure for the test run.


totten:
Thanks, Anthony. This looks good.

PR #50, opened 2015-03-10
Drupal phpcs file says this is the correct comment format for implements

eileenmcnaughton:


eileenmcnaughton:
As an aside I also get complaints that 

'phpcs Hook implementation should not duplicate @param documentation' - although phpdoc inspection gets grumpy when the param is not there! Can't win.

eileenmcnaughton:
NB - what are your thoughts about adding some of the hooks functions commented out e.g

/**
 * Implements hook_civicrm_preProcess().
 *
 * @link http://wiki.civicrm.org/confluence/display/CRMDOC/hook_civicrm_preProcess
 *
function {modname}_civicrm_preProcess($formName, &$form) {

}
 */

Certainly makes it easy as a module developer to just be able to uncomment them

totten:
Patch looks good.

For the ```@param``` contention, we probably need to refer upstream to Drupal coder. (Or maybe merge / see if upstream still has issue.)

Some more stubs like that sound useful.

PR #52, opened 2015-03-27
Fix regex extracting hostname from base URL

antrik:
This fixes Drupal bootstrap for installs where the base URL is not the
root of the host. (Such as https://example.com/dcivi/ rather than just
https://example.com/ )


totten:
Thanks. I don't have this configuration, but I'll take your word for it that it works. ;)

PR #56, opened 2015-05-10
Improvements for invoking multiple test classes at once

antrik:
Two changes to facilitate running multiple test classes at once (by specifying a directory with tests):
  * don't change the working directory before running phpunit
  * default to tests/phpunit if no explicit class/directory has been requested

### DIFF
--- a/src/CRM/CivixBundle/Command/TestRunCommand.php
+++ b/src/CRM/CivixBundle/Command/TestRunCommand.php
@@ -31,7 +31,7 @@ class TestRunCommand extends ContainerAwareCommand {
     $this
       ->setName('test')
       ->setDescription('Run a unit test')
-      ->addArgument('<TestClass>', InputArgument::OPTIONAL, 'Test class name (eg "CRM_Myextension_MyTest")')
+      ->addArgument('<TestClass>', InputArgument::OPTIONAL, 'Test class name (eg "CRM_Myextension_MyTest"), or directory with test classes', 'tests/phpunit')
       ->addOption('clear', NULL, InputOption::VALUE_NONE, 'Clear the cached PHPUnit bootstrap data')
       ->addOption('filter', NULL, InputOption::VALUE_REQUIRED, 'Restrict tests by name (regex)')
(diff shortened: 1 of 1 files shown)


antrik:
I just realised that the working directory was actually important, because the extension directory path was relative. (Not sure why I didn't see this before...)

I added another commit now, to avoid the relative path.

totten:
I like the idea of being able to run folders of tests.

The observation that extensionsDir == 'extensions' is a bit weird. FWIW, on my system, I get an absolute path:

```
$ drush ev 'civicrm_initialize(); var_dump(\CRM_Core_BAO_Setting::getItem("Directory Preferences", "extensionsDir"));'
string(50) "/Users/totten/buildkit/build/d46/sites/default/ext"
```

I think that all the path settings (including ```extensionsDir```) are specified to be *either* an absolute path *or* a relative path. (Absolute paths can work in more diverse configurations, but relative paths are easier to migrate among servers/staging/prod/dev.)

In core's admin UI, the settings are always displayed as absolute paths. There's some filtering for both directions -- ie [```CRM_Core_BAO_Setting::retrieveDirectoryAndURLPreferences```](https://github.com/civicrm/civicrm-core/blob/4.6/CRM/Core/BAO/Setting.php#L1051) calls (CRM_Utils_File::absoluteDirectory)[https://github.com/civicrm/civicrm-core/blob/4.6/CRM/Utils/File.php#L556], and ```CRM_Core_BAO_Setting::storeDirectoryOrURLPreferences``` calls ```CRM_Utils_File::relativeDirectory```.

I think the current patch (```"$civicrm_root/tools/" . \CRM_Core_BAO_Setting::getItem('Directory Preferences', 'extensionsDir')```) would break in the situations where ```extensionsDir``` contains an absolute path. Perhaps it would make sense to pass paths through the same filter as core (eg ```CRM_Utils_File::absoluteDirectory```)?

antrik:
Indeed, this looks much better :-) In fact, loading extensions from the "proper" extension folder works for me now. (You might remember that I was wondering about this...)

I decided to submit this as a separate PR though, as this is really a totally unrelated issue -- it was pure coincidence that setting the working directory to "$civicrm_root/tools" was obscuring the path handling issue in my system...

I've restored this PR to its original state, as that was correct all along, and the other fix attempt was totally wrong.

PR #57, opened 2015-05-13
Fix handling of relative extension directory path

antrik:
Use CRM_Utils_File::absoluteDirectory() (as suggested by @totten), to
make sure the extension path is set to a correct value, even if
CRM_Core_BAO_Setting::getItem('Directory Preferences', 'extensionsDir')
returns a relative path.

This way, extension loading from the "proper" extension directory
actually works in my environment :-)

### DIFF
--- a/src/CRM/CivixBundle/Command/TestRunCommand.php
+++ b/src/CRM/CivixBundle/Command/TestRunCommand.php
@@ -179,7 +179,7 @@ class TestRunCommand extends ContainerAwareCommand {
       // disable extension searching
       $template_vars['civicrm_setting']['Extension Preferences']['ext_repo_url'] = FALSE;
       // use the same source tree for linked Civi runtime and test Civi runtime
-      $template_vars['civicrm_setting']['Directory Preferences']['extensionsDir'] = \CRM_Core_BAO_Setting::getItem('Directory Preferences', 'extensionsDir');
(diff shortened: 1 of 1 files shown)


totten:
Belated thanks for submitting the patch, although things have changed.

Basically, executing tests via `civix test` is deprecated -- I don't think it's possible for `civix test` to abstract-away the bootstrap issues involved with testing. Rather, these are a fundamental part of how you define a suite -- and different suites have different needs. (Ex: `CRM_AllTests` vs `WebTest_AllTests` vs `E2E_AllTests` vs any given extension.) It's better to let extension authors define their own suites -- and add-in bits of Civi functionality as needed.

A big piece to "mix-in" is the bootstrap, which can be handled via https://github.com/civicrm/cv , and this is demonstrated in https://github.com/civicrm/org.civicrm.testapalooza .

antrik:
@totten well, you may consider it deprecated -- but I'm still using it; and unless you actually removed that code path entirely (I hope not?), there is no reason to refuse merging the fix...

I agree that creating a reliable test environment with `civix test` is a pain -- but mostly just because CiviCRM's test suite bootstrap is such a horrible horrible mess; and to a lesser degree, because the way civix hooks into it is not ideal. Still, with a bunch of hacks, I actually managed to create a perfectly reliable environment.

I don't even understand half of what you wrote above. It sounds like considerable effort to learn this stuff, and to migrate to it. I'm not exactly motivated to abandon my working setup, unless I can see substantial advantages over what I have now...

PR #59, opened 2015-09-25
URL corrections

xurizaemon:
Trivial but helpful fixes


totten:
Thanks.

That's a pretty weird mistake. Kinda wonder if the page was moved or if it was copy-paste-edit mistake, but it's pretty academic. New links are better.

xurizaemon:
Yeah I think the DOC sprouted at some time in recent history, remember thinking it was odd.

¯\_(ツ)_/¯ yay fixed

PR #79, opened 2016-09-10
CRM-19252: Extension versions should be domain-independent

GinkgoFJG:


totten:
Looks nicely done. :) Thanks, @GinkgoFJG.

Tested a few edge-cases (e.g. clean installation; followed by patch to introduce schema; followed by patch to revise schema; then uninstallation; then reinstallation at the latest schema), and it seemed to work consistently.

Merged manually because I needed to update `UPGRADE.md` to reflect the merge date.

PR #85, opened 2016-12-23
instruct on non sudo and settings path setup

ErichBSchulz:
This seems to be working for me.

(I haven't tested it thoroughly but generate:module certainly works!

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ Start Menu -> Control Panel -> System -> Advanced -> Environment Variables
 
 ### Environment
 
-`civix` will pick up your name and email from you git installation if it can detect this.
+`civix` will pick up your name and email from your git installation if it can detect this.
 
 It will also look for your `civicrm.settings.php` file. If it cannot find the file then set the path to it:
 
--- a/README.md
+++ b/README.md
@@ -12,9 +12,18 @@ Civix is a command-line tool for building CiviCRM extensions.
 Civix is distributed as a single, portable PHAR executable.  As long as you have PHP-CLI
 properly configured, it should work as a simple download, e.g.
 
+To install globally:
+
 ```bash
 sudo curl -LsS https://download.civicrm.org/civix/civix.phar -o /usr/local/bin/civix
 sudo chmod +x /usr/local/bin/civix
+```
+
+To install as a user without `sudo`:
+
+```bash
+curl -LsS https://download.civicrm.org/civix/civix.phar -o ~/bin/civix
+chmod +x ~/bin/civix
 ```
 
 To upgrade an existing installation, simply re-download the latest `civix.phar`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,11 +12,20 @@ Civix is a command-line tool for building CiviCRM extensions.
...
 properly configured, it should work as a simple download, e.g.
 
+To install globally:
+
 ```bash
 sudo curl -LsS https://download.civicrm.org/civix/civix.phar -o /usr/local/bin/civix
 sudo chmod +x /usr/local/bin/civix
 ```
 
+To install as a user without `sudo`:
```
totten: @ErichBSchulz I appreciate the sentiment this, and personally I don't actually use `sudo` for installing `civix`, but...

As a publisher of PHP CLI tools, my experience in documenting CLI installation is that's a bit sisyphean if you try to anticipate the different permutations. In some contexts, `/usr/local/bin` works; or maybe the home directory or `/opt/<foo>/bin` is better. Some people don't understand what a home-directory is for -- or that `~` refers to it. Often the chosen directory isn't in the PATH, so you need to edit a config-file or system-setting, but the filename or setting differs by OS/distro/shell. (Even just with `bash`, you can choose `~/.bashrc` or `~/.profile` or `~/.bash_profile` or some equivalent under `/etc`). Or you need other steps besides `chmod +x` (like creating a `.bat` wrapper on Windows). Or the user doesn't have `curl` but does have `wget` or a graphical web-browser.

Philosophically, I don't think each CLI tool should reproduce install-docs for these variations. The steps are the same for all the PHAR tools (`drush` or `wp-cli` or `phpunit` or `civix` or  `cv` or `civistrings` or  `amp` or `git-scan`) -- the differences which arise are environmental/contextual/subjective. I'd prefer to just publish the executable and assume that the user already knows how to handle basic admin on their own computer.

But pragmatically we can't provide *zero* install instructions, and many new users come in with weak sysadmin skills...

The current example for global installation is a compromise: it's the shortest list of steps I could find that would work out of the box in a broad swath of Unix-y systems (e.g. Ubuntu+OS X) -- and if a user understands his system well enough to prefer something else, then he should be able to improvise based on these minimal steps.

I think the PHP ecosystem is missing a particular resource -- a generic installation guide for PHP CLI tools that covers 90% of environments. This would explain things like different PHP runtimes, home dirs, shell config, PHAR, and `composer global` -- as typically applied in OS X, Windows, and Debian/Ubuntu. The `README.md` for `civix.phar` or `cv.phar` or whatever would then include one basic example -- plus a link to the general guide.

README.md
```diff
@@ -67,6 +76,14 @@ OR permanently:
 Start Menu -> Control Panel -> System -> Advanced -> Environment Variables
 ```
 
+### Environment
```
totten: I think these are good things to point out.

PR #86, opened 2016-12-23
Automate detection of new entity types

MegaphoneJon:
Running `civix generate:entity` creates a `<name>.entityType.php` file very similar in structure to the `.mgd.php` files, but doesn't appear to be automatically detected similarly.  This patch eliminates the need to manually define `hook_civicrm_entityTypes`.

### DIFF
--- a/src/CRM/CivixBundle/Resources/views/Code/module.civix.php.php
+++ b/src/CRM/CivixBundle/Resources/views/Code/module.civix.php.php
@@ -200,6 +200,24 @@ function _<?php echo $mainFile ?>_civix_civicrm_managed(&$entities) {
 }
 
 /**
+ * (Delegated) Implements hook_civicrm_entityTypes().
+ *
+ * Find any *.entityType.php files, merge their content, and return.
+ *
+ * @link http://wiki.civicrm.org/confluence/display/CRMDOC/hook_civicrm_entityTypes
+ */
+function _<?php echo $mainFile ?>_civix_civicrm_entityTypes(&$entityTypes) {
+  $entityTypeFiles = _<?php echo $mainFile ?>_civix_find_files(__DIR__, '*.entityType.php');
+  foreach ($entityTypeFiles as $file) {
+    $ets = include $file;
+    foreach ($ets as $et) {
+      $name = $et['name'];
+      $entityTypes[$name] = $et;
+    }
+  }
+}
+
+/**
  * (Delegated) Implements hook_civicrm_caseTypes().
  *
  * Find any and return any files matching "xml/case/*.xml"
--- a/src/CRM/CivixBundle/Resources/views/Code/module.php.php
+++ b/src/CRM/CivixBundle/Resources/views/Code/module.php.php
@@ -98,6 +98,17 @@ function <?php echo $mainFile ?>_civicrm_managed(&$entities) {
 }
 
 /**
+ * Implements hook_civicrm_entityTypes().
+ *
+ * Declare entity types provided by this module.
+ *
+ * @link http://wiki.civicrm.org/confluence/display/CRMDOC/hook_civicrm_entityTypes
+ */
+function <?php echo $mainFile ?>_civicrm_entityTypes(&$entityTypes) {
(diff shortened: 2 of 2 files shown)


PalanteJon:
Hmm - now I'm thinking I'm not sure I completely understand how the caching works for entity types; I thought of this as being more similar to `hook_civicrm_manage` than it probably is.  Any thoughts Tim?

totten:
> Hmm - now I'm thinking I'm not sure I completely understand how the caching works for entity types; I thought of this as being more similar to `hook_civicrm_manage` than it probably is. Any thoughts Tim?

@PalanteJon Yup, that's the nub. This patch is perfect in terms of carrying `civix` coding conventions to `hook_civicrm_entityTypes`, but that hook has different performance implications than `hook_civicrm_managed`.

By design, managed entities are only examined on rare occasion (during a system flush), so it's not a huge problem if `hook_civicrm_managed` does a file-system scan (slowish operation) via `_mymod_civix_find_files()` or  `_mymod_civix_glob()`. The same is true of file-scans in `hook_civicrm_xmlMenu` and `hook_civicrm_caseTypes`.

However, the list of entity/dao/table mappings are used frequently. (Possibly every API call?) Calling `*_civix_find_files(...)` in `hook_civicrm_entityTypes` would mean that (effectively) every extension does a file-system search on every page-request. 

There is a cache in  `CRM_Core_DAO_AllCoreTables`, but it's `static`. That reduces lookups within the same page-request but doesn't address the problem above. If `CRM_Core_DAO_AllCoreTables` used a persistent cache, then that would probably address it?

totten:
@MegaphoneJon  - @michaelmcandrew asked about this topic in the discussion of #124.

Recapping, the basic concern here was, "How do we avoid performing a file-system scan on every page request?" 

I wonder if this simple change would mitigate that enough:

```php
function _<?php echo $mainFile ?>_civix_civicrm_entityTypes(&$entityTypes) { 
  $schemaDir = implode(DIRECTORY_SEPARATOR, [__DIR__, 'xml', 'Schema']);
  if (!file_exists($schemaDir)) return;

  $entityTypeFiles = _<?php echo $mainFile ?>_civix_find_files($schemaDir, '*.entityType.php');
  foreach ($entityTypeFiles as $file) ...
}
```

That significantly reduces of the number of lookups by (a) failing quicker on typical extensions and (b) focusing on a smaller part of the tree. Example: suppose there are 25 extensions, including 5 that have their own entities; assume an average of 25 inodes per extension, of which maybe 5 are in `xml/Schema`. The change reduces the number of file-checks from `25*25==625` to `25+(5*5)=50`.

But 50 still feels a bit high for something that happens on every page-request. It may be worthwhile to recap some options:

 * *Narrow the scope of `_civix_civicrm_entityTypes`* (as above). Simple change. However, the  execution-time still has some multiplication (`O(#extensions + #daoExtensions*#xmlInodes)`).
 * *Patch `civicrm-core` to cache `hook_civicrm_entityTypes`.* On a typical page-view, this brings the cost down to constant-time (one lookup per page-view). However,  it's only going to help on upgraded sites; and it may or may not be easy (`hook_civicrm_entityTypes` fires very early during the bootstrap process).
 * *Patch `civix generate:entity` to put all metadata in one file* (e.g. `mymodule.entityTypes.php` or `mymodule.entityTypes.json`). On a typical page-view, this would bring it down to linear-time (one lookup per extension).
 * *Continue prompting developers to manually setup the hook.*

michaelmcandrew:
Hey both,

Another approach since performance is important here:

Have `generate:entity-boilerplate` scan for these files and generate an _civix_civicrm_entityTypes function based on what it finds.

michaelmcandrew:
Hey @totten and @MegaphoneJon - I ended up implementing an alternative approach to this problem here: https://github.com/totten/civix/pull/124#issuecomment-368538364 - got it down to 0 lookups :) 

Let me know what you think.

michaelmcandrew:
@totten and @MegaphoneJon - we can probably close this issue now.

totten:
Agree. Thanks @MegaphoneJon @michaelmcandrew .

PR #87, opened 2016-12-24
add some comments and clarify script actions

ErichBSchulz:
hope this helps!


totten:
Thanks, @ErichBSchulz! The help message seems better and the WORKINGDIR is good.

I made a couple small tweaks (e.g. 4eb7a4c5 and fc25e73d).

PR #88, opened 2016-12-27
Add title to angular page

colemanw:


colemanw:
@totten is it ok to do a quick pr like this or does code need recompiling?

totten:
@colemanw In this case, I think a quick PR makes sense. I don't think there are any other pieces which depend on `angular-page.html.php`, so it should be pretty safe to revise that template (e.g. no need to notify downstream devs).

Jenkins (https://test.civicrm.org/job/Tool-Publish-civix/) will pickup on changes in the `civix` repo and update the `civix.phar` file on `download.civicrm.org`  . However, buildkit pegs most CLI tools to specific versions -- https://github.com/civicrm/civicrm-buildkit/blob/master/bin/civi-download-tools#L20 -- so that's a separate thing.

PR #93, opened 2017-01-01
prevent untidy warning when xpath missing

ErichBSchulz:
according to symfony docs its impossible to have a "required" option

(if user specifies the option then it is possible to require a value)

the way around this seems to be to use an "argument"...

I'm assuming that we wouldn't be keen to alter the signature now - but this patch at leasts give more useful output (i was getting a couple of pages of stack trace with `civix info:get `)

hope this is helpful!

### DIFF
--- a/src/CRM/CivixBundle/Command/InfoGetCommand.php
+++ b/src/CRM/CivixBundle/Command/InfoGetCommand.php
@@ -37,7 +37,7 @@ class InfoGetCommand extends Command {
     $this
       ->setName('info:get')
       ->setDescription('Read a field from the info.xml file')
-      ->addOption('xpath', 'x', InputOption::VALUE_REQUIRED, '(REQIRED) The XPath expression of the field')
+      ->addOption('xpath', 'x', InputOption::VALUE_REQUIRED, '(REQUIRED) The XPath expression of the field')
       ->setHelp("Read a single field from the info.xml file.
 
 Examples:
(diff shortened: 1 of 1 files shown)


src/CRM/CivixBundle/Command/InfoGetCommand.php
```diff
@@ -36,7 +37,7 @@ protected function configure() {
     $this
       ->setName('info:get')
       ->setDescription('Read a field from the info.xml file')
-      ->addOption('xpath', 'x', InputOption::VALUE_REQUIRED, 'The XPath expression of the field')
+      ->addOption('xpath', 'x', InputOption::VALUE_REQUIRED, '(REQIRED) The XPath expression of the field')
```
totten: Typo

src/CRM/CivixBundle/Command/InfoGetCommand.php
```diff
@@ -54,7 +55,13 @@ protected function execute(InputInterface $input, OutputInterface $output) {
     $info->load($ctx);
     $info->get();
 
-    foreach ($info->get()->xpath($input->getOption('xpath')) as $node) {
+    $xpath = $input->getOption('xpath');
+    if (is_null($xpath)) {
+      $help = new HelpCommand();
```
totten: 1. Is this something you can do by tweaking the `addOption(...)`  data to make the field required?

2. When delegating to subcommands, use something like https://github.com/joomlatools/joomlatools-console/commit/019c5cbfc8c601953562f3a097893a1eeac216ef

totten:
@ErichBSchulz There are two issues that need fixing in this PR.

src/CRM/CivixBundle/Command/InfoGetCommand.php
ErichBSchulz: sorry Tim

apparently ["Options are always optional,"](http://symfony.com/doc/current/console/input.html)...

ErichBSchulz:
thank Tim - have made those tweaks

this is a bit of an imperfect fix as changing from an option to an argument would require a change in the command signature and that seems a bit excessive

totten:
Thanks! Looks good now.

PR #94, opened 2017-01-25
define managed entities with api3 by default

michaelmcandrew:
At this point in time, I think it makes sense to presume that managed entities are defined using api3.

Please merge if you agree :)


totten:
Yup, this sounds right.

PR #96, opened 2017-05-31
generate:angular-page - improve help text

seanmadsen:
Clarify that an Angular module must be generated first
https://civicrm.stackexchange.com/questions/18837/unknown-path-when-viewing-a-newly-generated-angular-page


totten:
Good point.

PR #99, opened 2017-07-28
Make it clear that civix test is deprecated

mattwire:


totten:
Thasnk @mattwire. We do want this to be clear.

FWIW, the newer recommended practice was documented in https://github.com/civicrm/org.civicrm.testapalooza

Issue #100, opened 2017-08-08
Undefined variable: path in CRM_Optivo_ExtensionUtil::path()


tschuettler:
Undefined variable: path
/opt/buildkit/build/dmaster/sites/default/files/civicrm/ext/de.oxfam.optivo.connector/optivo.civix.php:56

totten:
Closed by merging #101 101.

PR #101, opened 2017-08-08
#100 fix undefined variable: path

tschuettler:


totten:
Thanks, @tschuettler . Good catch!

PR #102, opened 2017-08-09
Extend generate:module command to regenerate module.civix.php

mattwire:


totten:
Thanks! Tried it out and it works for me. :) This looks more ergonomic than the current advice in`UPGRADE.md` -- it's easier to internalize / develop some muscle-memory if you don't have to retype the extension name. 

Before merging, I would like your feedback on somewhat bikesheddy point -- could we name the command something else?
 * Disclosure: the naming convention for `civix` is a bit quirky IMHO (as xurizaemon has pointed out before.... although I place the blame with Symfony's `console` from whence `civix` was born...). Specifically, code-generators are named  `generate:<noun>`, but other commands are mostly named `<noun>:<verb>` or maybe `<subsystem>:<verb>`. I'm not sure the convention encourages good name here.
 * The name `generate:refresh` gets us into `<verb>:<verb>` territory. 
 * The action of "refreshing" can actually apply to other artifacts (like the base-class for the upgrader).

Some alternatives to kick around...
 * Rename to `civix refresh:module`
 * Rename to `civix refresh` -- a general upgrade command (that might scan and update multiple artifacts)
 * Revise `civix generate:module` to handle this better -- eg make the extension-name optional. If it's blank and if there's an `info.xml`, use that.

The last one is probably my favorite, but I'm not super-opinionated about it. If you're still inclined toward `generate:refresh` (or something less `<verb>:<verb>`-y), then we can do that. Just comment or update PR to taste.

mattwire:
Hi, wasn't too sure about the name, but then I realised I just wanted something that worked so didn't worry too much!
`civix refresh` is perhaps more discoverable (ie. you find it when you ask for a list of commands) but `civix generate:module` would also be a good place for it.  And yes it would probably be nice to update other bits of code too (like the upgrader).

totten:
OK. Could you do one of those changes? Or we should leave it as an open TODO?

mattwire:
@totten Try that :-)

mattwire:
@totten Are you happy with these changes - happy to merge?

PR #105, opened 2017-10-08
Add "generate:test --template=phpunit" for minimalist test

pbatroff:
PR for general PHP Unittest integration in Civix
  * Add template 'phpunit'
  * Add example test to civix test script

### DIFF
--- a/src/CRM/CivixBundle/Command/AddTestCommand.php
+++ b/src/CRM/CivixBundle/Command/AddTestCommand.php
@@ -32,6 +32,9 @@ In creating a test, you may specify a template:
             effort.
   legacy:   A variation of `headless` based on CiviUnitTestCase.
             It is provided primarily for testing purposes.
+  phpunit:  A test suite based on the PHPUnit_Framework_TestCase. Provides an 
+            interface for you extension to implement unittest for your 
+            classes/functions
 
 To execute tests, call phpunit 4.x directly, e.g.
 
@@ -104,6 +107,7 @@ as separate groups:
       'e2e' => 'test-e2e.php.php',
       'headless' => 'test-headless.php.php',
       'legacy' => 'test-legacy.php.php',
+      'phpunit' => 'test-phpunit.php.php',
     );
     if (isset($templates[$type])) {
       return $templates[$type];
--- a/src/CRM/CivixBundle/Resources/views/Code/test-phpunit.php.php
+++ b/src/CRM/CivixBundle/Resources/views/Code/test-phpunit.php.php
@@ -1,0 +1,41 @@
+<?php
+echo "<?php\n";
+?>
+
+// Allow autoloading of PHPUnit helper classes in this extension.
+$loader = new \Composer\Autoload\ClassLoader();
+$loader->add('CRM_', __DIR__);
+$loader->add('Civi\\', __DIR__);
+$loader->register();
+
+/**
+* This is a generic test class for you extension, implementing PHPUnit test
+*
+*   TODO: Add more description/explanation in here
+*/
+class <?php echo $testClass ?> extends \PHPUnit_Framework_TestCase {
+
+    /**
+     * Setup Method is executed before the test is executed
+     */
+    public function setUp() {
+        parent::setUp();
+    }
+
+    /**
+     * Tear Down Method is executed after the test was executed
+     * This should be used for cleanup
+     */
+    public function tearDown() {
+        parent::tearDown();
+    }
+
+    /**
+     * Example test case
+     * TODO: Description
+     */
+    public function testExample() {
+        $myBool = True;
+        self::assertTrue($myBool, "The argument must be true to pass the test");
+    }
+}
--- a/src/CRM/CivixBundle/Resources/views/Code/test-phpunit.php.php
+++ b/src/CRM/CivixBundle/Resources/views/Code/test-phpunit.php.php
@@ -9,34 +9,32 @@ $loader->add('Civi\\', __DIR__);
 $loader->register();
 
 /**
- * This is a generic test class for the extension, implementing PHPUnit tests.
- *
- * Testing functions must begin 'test' to be included in the PhpUnit tests.
+ * This is a generic test class for the extension (implemented with PHPUnit).
  */
 class <?php echo $testClass ?> extends \PHPUnit_Framework_TestCase {
 
     /**
-     * Setup Method is executed before the test is executed (optional)
+     * The setup() method is executed before the test is executed (optional).
      */
     public function setUp() {
         parent::setUp();
     }
 
     /**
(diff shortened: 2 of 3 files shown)


src/CRM/CivixBundle/Command/AddTestCommand.php
```diff
@@ -32,6 +32,9 @@ protected function configure() {
             effort.
   legacy:   A variation of `headless` based on CiviUnitTestCase.
             It is provided primarily for testing purposes.
+  phpunit:  A test suite based on the PHPUnit_Framework_TestCase. Provides an
```
mickadoo: you extension => your extension
unittest => unit testing

src/CRM/CivixBundle/Resources/views/Code/test-phpunit.php.php
```diff
@@ -0,0 +1,42 @@
+<?php
+echo "<?php\n";
+?>
+
+// Allow autoloading of PHPUnit helper classes in this extension.
+$loader = new \Composer\Autoload\ClassLoader();
+$loader->add('CRM_', __DIR__);
+$loader->add('Civi\\', __DIR__);
```
mickadoo: @totten 

I know I'm harping on about PSR-4, but I don't see why we shouldn't advocate it in new extensions since it's supported. 

I could be wrong, but if we're using the Civi namespace root suggested above I think we'd end up with this directory structure to namespace mapping:

Path: `{extension.root}/tests/Civi/Test/MyCompany/MySuperExtension/Services/MyClassTest.php` 
ClassName: `Civi\Test\MyCompany\MySuperExtension\Services\MyClassTest.php`

And if we went for PSR-4:

`$loader->addPsr4('Test\MyCompany\MySuperExtension\', __DIR__);`
We could remove a few levels:

Path: `{extension.root}/tests/Services/MyClassTest.php `
Namespace: `Test\MyCompany\MySuperExtension\Services\MyClassTest.php`

src/CRM/CivixBundle/Resources/views/Code/test-phpunit.php.php
```diff
@@ -0,0 +1,42 @@
...
+        parent::tearDown();
+    }
+
+    /**
+     * Example test case
+     *  Simple test if a boolean expression is true. This is implemented to show
+     *  an example for a test function
+     */
+    public function testExample() {
+        $myBool = True;
```
mickadoo: Since this is going to be deleted anyway I'd suggest just limiting it to one line, also according to Civi coding standards the booleans should be in uppercase.

I know it's going to be changed anyway, but chances are new people seeing this test will copy it. Same goes for the assertion statement. There is no [right or wrong way](https://phpunit.de/manual/current/en/appendixes.assertions.html#appendixes.assertions.static-vs-non-static-usage-of-assertion-methods) but personally I use `$this->assertTrue` more than static calls.

src/CRM/CivixBundle/Resources/views/Code/test-phpunit.php.php
totten: PSR-4 is great, but starting it in this context would leave the system is a rather confusing state -- because it would encourage downstream to use a mix of different file-structures *within the same extension*.

Rather, consider adding a flag to specify that an extension should use namespaces and flat directory trees, e.g.

  * When creating a module pass: `civix generate:module org.example.foo --layout=psr-4`
  * In the `info.xml`, store: `<civix><file-structure>psr-4</file-structure</civix>`
  * In subcommands like `generate:test`, `generate:upgrader`, and `generate:page`, it should create files based on that convention.

If we started out by *just* changing the `generate:test` to hard-code a PSR-4 layout, then we'd be encouraging everyone to adopt a weird hybrid layout.

This is not to say that switching to PSR-4 needs to be done all-at-once/perfectly. But it is a big change, and I wouldn't want this to become *that one quirky thing* that stems from an abortive/half-hearted design-change. The opening move in a change toward PSR-4 would need to be (a) adding the new flag and (b) using that flag in a couple of commands/templates.

totten:
@pbatroff - ditto on the textual copy-edits suggested by @mickadoo . Also, I updated your branch with some more tweaks to the prose (d7240ef).

One general issue that needs patching -- it looks like it uses a 4-space indent, which doesn't comply with https://docs.civicrm.org/dev/en/latest/standards/php/ . You can test/see this as follows:

```bash
me@localhost:~/buildkit/build/dmaster/sites/all/modules/civicrm/ext/org.example.foo$ ~/src/civix/bin/civix generate:test --template=phpunit CRM_Foo_BarTest
Write /Users/totten/buildkit/build/dmaster/sites/all/modules/civicrm/ext/org.example.foo/phpunit.xml.dist
Write /Users/totten/buildkit/build/dmaster/sites/all/modules/civicrm/ext/org.example.foo/tests/phpunit/bootstrap.php
Write /Users/totten/buildkit/build/dmaster/sites/all/modules/civicrm/ext/org.example.foo/tests/phpunit/CRM/Foo/BarTest.php
me@localhost:~/buildkit/build/dmaster/sites/all/modules/civicrm/ext/org.example.foo$ civilint  tests/phpunit/CRM/Foo/BarTest.php
===========================[ Identify Files ]===========================
PHP Files:
 * tests/phpunit/CRM/Foo/BarTest.php
===============================[ php -l ]===============================
===============================[ phpcs  ]===============================

FILE: ...l/modules/civicrm/ext/org.example.foo/tests/phpunit/CRM/Foo/BarTest.php
--------------------------------------------------------------------------------
FOUND 5 ERRORS AFFECTING 5 LINES
--------------------------------------------------------------------------------
 14 | ERROR | [x] Line indented incorrectly; expected 2 spaces, found 4
 17 | ERROR | [x] Line indented incorrectly; expected 2 spaces, found 4
 18 | ERROR | [x] Line indented incorrectly; expected 4 spaces, found 8
 26 | ERROR | [x] Line indented incorrectly; expected 6 spaces, found 8
 35 | ERROR | [x] Line indented incorrectly; expected 6 spaces, found 8
--------------------------------------------------------------------------------
PHPCBF CAN FIX THE 5 MARKED SNIFF VIOLATIONS AUTOMATICALLY
--------------------------------------------------------------------------------

Time: 132ms; Memory: 5.25Mb

Found PHP errors
```

src/CRM/CivixBundle/Resources/views/Code/test-phpunit.php.php
mickadoo: Fair point on the appropriateness of making the change here. I agree that it is probably best to avoid approaching the change as part of a big single PR where everything is changed to PSR-4, not many people are going to volunteer for that one.

Flags and conditionals here seem like a bit of over-complication. I think this is where we have a single approach which is the recommended best practice by CiviCRM and document the other options to allow people to choose what they like.

PR #106, opened 2017-10-08
When creating a new API, automatically create a unit test.

pbatroff:
PR to automatically add unittest file generation upon generating an API command

### DIFF
--- a/src/CRM/CivixBundle/Builder/PHPUnitGenerateInitFiles.php
+++ b/src/CRM/CivixBundle/Builder/PHPUnitGenerateInitFiles.php
@@ -1,0 +1,45 @@
+<?php
+namespace CRM\CivixBundle\Builder;
+use CRM\CivixBundle\Services;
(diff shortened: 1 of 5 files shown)


totten:
@pbatroff Looking like progress. :)

A couple notes:
 * In the docblocks, the `*`'s aren't aligned.
 *  For the content of the test, let's do include a call to an API. Maybe like this:
   ```php
   $result = civicrm_api3('<?php echo $entityNameCamel ?>', '<?php echo $actionNameCamel ?>', array('magicword' => 'sesame'));
   $this->assertEquals('Twelve', $result['values'][12]['name']);
   ```
 * Testing the API will require more services from Civi, so it look more like the `headless` template -- ie
    * Add several `use Civi...` statements to include some interfaces
    * The `class` should `implement` those interfaces
    * The `setupHeadless()` function should initialize the runtime environment.

PR #111, opened 2017-10-11
Integrate codeception to Civix

pbatroff:
See also: https://github.com/civicrm/civicrm-core/pull/11115


pbatroff:
Any chance to get this work from the sprint integrated? Would be really nice to have this as an option; @totten Please let me know if anything needs fixing, so I can allocate some time! 

Thanks Phil

PR #115, opened 2017-12-20
Short array syntax - woohoo!

colemanw:


colemanw:
@totten this updates the php templates for auto-generated files as well as the base code.
Since it's a big patch it will be prone to merge conflicts. Please merge when you can.

totten:
Thanks @colemanw. Reading the patch, it looks pretty good. (Couldn't spot any problems - though admittedly it's long.)

I've split the PR into a couple smaller chunks:

* #117 (civix logic) - I'm reasonably comfortable with merging this part. The test-process gives a straight-forward pass, and the change only escalates PHP requirements on developer boxes. It's also the larger portion of the changes, so merging it should cut back the risk of MC's.
* #118 (civix templates) - I'm a little less comfortable with this part. The test-process doesn't give quite such a clear 👍 or 👎, and the change will cause downstream extensions to require PHP 5.4+. Did you (by chance) do testing with this part (which might not be apparent in the PR/patch)?

PR #122, opened 2018-02-08
Allow simpler (non java like) extension names with dashes

michaelmcandrew:
Previously, civix required / encouraged extension authors to follow java naming conventions in their extensions and include an org name as part of the extension name. This PR removes that requirement.

Extension names can now include `-` and do not require a `.`

If one or more `.`s are found in the extension name, a short name is generated from the characters after the last `.`, else the name and short name are equivalent.

Also, good news for people that like to start their extension names with with `civicrm-`. Civix now trims `civicrm-` from the start of the short name when found before creating the file and function names and the namespace.

The old style of extension naming is still accepted.

Try generating the following extensions to get an idea of how it works

```
civix generate:module a
civix generate:module a.b
civix generate:module a.b.c
civix generate:module alta-vista
civix generate:module civicrm-do-thing
civix generate:module org.example.civicrm-apple-crumble
civix generate:module org.example.vanilla-ice
```


totten:
@michaelmcandrew I've tightened the name validation (e.g. prevent `foo...org` or `foo-bar--`), added some tests, and added some prose for help text.

PR #124, opened 2018-02-22
new command to generate boilerplate from XML schema file

michaelmcandrew:
Hey there,

Automatically creating DAO files and the associated SQL statements for install / uninstall has been missing from civix for a while. It is a pain to do this manually: (step 4 of https://docs.civicrm.org/dev/en/latest/extensions/civix/#generate-entity).

This PR adds a new command `civix generate:daos` to automate that step.

The expected workflow is that you have run `civix generate:entity` 1 or more times to generate entities, and have edited the schema xml files to your liking. You then run `civix generate:daos` which generates DAOs and creates auto install / uninstall sql for the extension as a whole.

The approach I took was to reuse the bare minimum from `CRM_Core_CodeGen_*` rather than try do it the civix way. It is admitedly a bit hacky, and I would understand if there were reasons to not include it in civix, but on the other hand, it is better than nothing.

I have added `(*EXPERIMENTAL AND INCOMPLETE*)` to the command to enourage you to accept the PR :)

### DIFF
--- a/src/CRM/CivixBundle/Command/AddDaosCommand.php
+++ b/src/CRM/CivixBundle/Command/AddDaosCommand.php
@@ -20,7 +20,7 @@ class AddDaosCommand extends \Symfony\Component\Console\Command\Command {
   protected function configure() {
     $this
       ->setName('generate:daos')
-      ->setDescription('Create DAOs and mysql schema based on this extensions xml schema definition files')
+      ->setDescription('Create DAOs and mysql schema based on this extensions xml schema definition files (*EXPERIMENTAL AND INCOMPLETE*)')
       ->setHelp(
         "Create DAOs and mysql schema based on this extensions xml schema definition files\n" .
         "\n" .
--- a/src/CRM/CivixBundle/Command/AddDaosCommand.php
+++ b/src/CRM/CivixBundle/Command/AddDaosCommand.php
@@ -1,0 +1,104 @@
+<?php
+namespace CRM\CivixBundle\Command;
+
+use CRM\CivixBundle\Services;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use CRM\CivixBundle\Builder\Collection;
+use CRM\CivixBundle\Builder\Dirs;
+use CRM\CivixBundle\Builder\Info;
+use CRM\CivixBundle\Builder\PhpData;
+use CRM\CivixBundle\Builder\Template;
+use CRM\CivixBundle\Utils\Path;
+use Exception;
+
+class AddDaosCommand extends \Symfony\Component\Console\Command\Command {
+  const API_VERSION = 3;
+
+  protected function configure() {
+    $this
+      ->setName('generate:daos')
+      ->setDescription('Create DAOs and mysql schema based on this extensions xml schema definition files')
+      ->setHelp(
+        "Create DAOs and mysql schema based on this extensions xml schema definition files\n" .
+        "\n" .
+        "Typically, one will have created / updated on or more xml/CRM/NameSpace/EntityName.xml files\n" .
+        "before running this command.\n"
+      );
+  }
+
+  protected function execute(InputInterface $input, OutputInterface $output) {
+    // load Civi to get access to civicrm_api_get_function_name
+    Services::boot(['output' => $output]);
+    $civicrm_api3 = Services::api3();
+    if (!$civicrm_api3 || !$civicrm_api3->local) {
+      $output->writeln("<error>Require access to local CiviCRM source tree. Configure civicrm_api3_conf_path.</error>");
+      return;
+    }
+
+    $ctx = [];
+    $ctx['type'] = 'module';
+    $ctx['basedir'] = \CRM\CivixBundle\Application::findExtDir();
+    $basedir = new Path($ctx['basedir']);
+
+    $info = new Info($basedir->string('info.xml'));
+    $info->load($ctx);
+    $attrs = $info->get()->attributes();
+    if ($attrs['type'] != 'module') {
+      $output->writeln('<error>Wrong extension type: ' . $attrs['type'] . '</error>');
+      return;
+    }
+
+    $xmlSchemaGlob = "xml/schema/{$ctx['namespace']}/*.xml";
+    $absXmlSchemaGlob = $basedir->string($xmlSchemaGlob);
+    // var_dump(($absXmlSchemaGlob));
+    $xmlSchemas = glob($absXmlSchemaGlob);
+
+    if(!count($xmlSchemas)){
+      throw new Exception("Could not find files matching '$xmlSchemaGlob'. You may want to run `civix generate:entity` before running this command.");
+    }
+
+    $specification = new \CRM_Core_CodeGen_Specification;
+    $specification->buildVersion = \CRM_Utils_System::majorVersion();
+
+    $config = new \stdClass;
+    $config->phpCodePath = $basedir->string('');
+    $config->sqlCodePath = $basedir->string('sql/');
+
+    foreach($xmlSchemas as $xmlSchema){
+
+      $dom = new \DomDocument();
+      $xmlString = file_get_contents($xmlSchema);
+      $dom->loadXML($xmlString);
+      $xml = simplexml_import_dom($dom);
+
+      $specification->getTable($xml, $database, $tables);
+
+      $tables[(string) $xml->name]['sourceFile'] = $xmlSchema;
+      $config->tables = $tables;
+
+      $dao = new \CRM_Core_CodeGen_DAO($config, (string) $xml->name);
+      ob_start(); // Don't display gencode's output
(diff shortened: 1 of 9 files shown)


src/CRM/CivixBundle/Command/AddDaosCommand.php
```diff
@@ -0,0 +1,104 @@
...
+      $xmlString = file_get_contents($xmlSchema);
+      $dom->loadXML($xmlString);
+      $xml = simplexml_import_dom($dom);
+
+      $specification->getTable($xml, $database, $tables);
+
+      $tables[(string) $xml->name]['sourceFile'] = $xmlSchema;
+      $config->tables = $tables;
+
+      $dao = new \CRM_Core_CodeGen_DAO($config, (string) $xml->name);
```
totten: I haven't `r-run` this PR yet, but -- based on skimming -- this looks like the real deal for generating DAOs in `civix`. Much nicer than hybrid bash/manual workflows we've used before...

totten:
## Quick trial

So I did a quick trial run, which basically meant:

```
civix generate:module org.example.foo
cd org.example.foo
civix generate:upgrader
civix generate:entity Foosball
civix generate:daos
```

And this worked. Installing the extension would create the SQL table; uninstalling would drop it. Hoowa!

## On "Experimental" flag

I do appreciate keeping the "EXPERIMENTAL" flag on it -- not because of any functional complaint with the code-generator (although I have some nitpicks below). The functionality seems generally usable (and much better than before).

The tension is more around DAOs generally: *In practice*, `dao.tpl` doesn't change much, so there aren't many problems with extensions distributing DAOs. But *in principle*, `dao.tpl` could change, and it would be difficult to coordinate an update among extensions. So I'd still encourage anyone using to exercise some vigilance wrt maintenance.

## Nitpicking

Consider the list of all deliverables for an entity:

 * `hook_civicrm_entityTypes` declaration
 * XML schema file
 * DAO file
 * BAO file
 * SQL create file
 * SQL drop file

It's a little incongruous that `generate:entity` doesn't produce a working entity, and it's a little incongruous that `generate:daos` produces more than DAOs (ie also `*.sql`). Rather, these commands seem to be more like steps in a workflow:

 * Step 1 - Generate editable templates
 * Step 2 - Edit the templates
 * Step 3 - Generate derivative boilerplate
 * (Optionally, repeat steps 2+3 to taste.)

At a bare minimum, the `generate:entity` command should probably display a message advising the user that `generate:upgrader` and `generate:daos` would be useful. But it does feel like some other change might make it a bit more intuitive.

michaelmcandrew:
Hey there,

Good that it worked for you.

> The tension is more around DAOs generally: In practice, dao.tpl doesn't change much, so there aren't many problems with extensions distributing DAOs. But in principle, dao.tpl could change, and it would be difficult to coordinate an update among extensions.

I'm not sure of the consequences of a change in the DAO template. I presume that DAOs generated by more that one template co-exist peacefully / we would want to avoid changing them in a way that prevented this from being true. Is the solution to package our own DAO template with civix?

The naming isn't brilliant, agreed - I didn't want to be too disruptive to what was currently in place. A couple of thoughts on how we could name stuff better...

`generate:entity` as it is currently called is 1-to-1 with entities. You can it once for every entity you want to generate.
`generate:daos` as it is currently called is 1-to-1 with the extension. You can think of it as the same as gencode for CiviCRM. Calling it `generate:code` would be somewhat in line with current practice, if a little vague :)

I'm leaning toward `generate:boilerplate` or `generate:entity-boilerplate`.

We could then, as you mentioned, remind people when running `generate:entity`, that they will need to run `generate:entity-boilerplate` at some point as well.

Since `generate:upgrader` is required to run the sql, we see if an upgrader has been defined and warn them if not.

[Just one more thing](https://www.youtube.com/watch?v=biW9BbWJtQU): should we be doing something (a_civix function or similar) to automate step 6 of https://docs.civicrm.org/dev/en/latest/extensions/civix/#generate-entity?

generate:entity seems to create something that can be used for that, but I don't think it gets registered. Or am I missing something?

totten:
Responding in reverse order, because that's a bit quicker....

`3.` For setting up the `hook_civicrm_entityTypes` to consume the `*.entityType.php` file, see #86.

`2.` 👍 for changes you described. Between `generate:boilerplate`, `generate:entity-boilerplate`, etc., I think they're all improvements -- happy to go with your gut.

`1.` For the consideration of `dao.tpl`, it's not enough to package another DAO template. Let me explain a little more.

If one makes a trivial whitespace change to the DAO template, and if there's mix of DAOs from old+new templates (that differ in whitepsace), then... that's fine. The potential problem comes from *changing the contract* for a DAO. Here are a couple examples that changed the contract:

 * https://github.com/civicrm/civicrm-core/commit/6b86d84f35a46b20e8414811f546edb38f3985e2#diff-4f4c04a1adf1c9b58ed78e1b0c2cf441R266 -- Adds a new function named `indices()`
 * https://github.com/civicrm/civicrm-core/commit/3cef3c00ce1e3de5ee93ee44bc1a1d1d3d406385#diff-4f4c04a1adf1c9b58ed78e1b0c2cf441L165 -- Updates `fields()` function to return more info

Both of these changes have the effect of *returning more information*. There are processes (like API Explorer; multilingual schema mgmt; etal) which consume that information. If the information is missing or inconsistent, then they may misbehave.

Of course, there's a lot of code that depends on DAO, so there's a lot of pressure to be conservative with changes. (I can't remember a time when someone *removed* metadata from DAOs.) And we now have at least one extension in the official test matrix with an external DAO. As far as I know, problems have been mostly theoretical so far. By contrast, the need for extension DAO's has been non-theoretical. I think it's a fair balance to add the generator and label it "Experimental".

If you want external DAOs to be on surer footing for the long-term, then that's really outside `civix.git`. The discussion probably belongs in a broader forum -- like https://lab.civicrm.org/dev/core or https://issues.civicrm.org/ .

michaelmcandrew:
Re: `3.` I suggested another approach in #86. I'll have a go at implementing that.

Re `1.` Thanks for the extra info - helps me to understand. I'll have a go at restating the problem. DAOs in extensions are sometimes consumed outside of the extension (API Explorer; multilingual schema mgmt). DAOs generated in our extension may not fulfil the contract that core is expecting of them.

Lets say core "adds stuff" to the DAO in 4.7.30, all we have to do to make an extension 4.7.30+ compatible is regenerate the DAOs and everything is compatible with <4.7.30 and >=4.7.30.

If core "changes / removes stuff" the solution is trickier, but not insolvable. We could make life easier for ourselves by only very rarely changing or removing stuff

So this is not a problem with `generate:entity-boilerplate`, but with having multiple DAO generators. And we have been recommending people do this in the docs for a while, so the problem is already in the wild.

The solution for now is adding version compatability notices. Longer term, the solution is something like do DAO generation as part of a post install build step.

michaelmcandrew:
@totten https://github.com/totten/civix/pull/124/commits/716e8e52ffe85ad7d54f20d89bd60516aacb479f collects entityTypes and populates a Civix function to define them

michaelmcandrew:
PS I tried the follow more complete routine:
```
civix generate:module org.example.clothes
cd org.example.clothes
civix generate:upgrader
civix generate:entity Hat
civix generate:entity-boilerplate
cv ext:uninstall clothes
cv ext:enable clothes
cv api Hat.create
cv api Hat.get
cv api Hat.eat
```
And it worked nicely.

michaelmcandrew:
See https://asciinema.org/a/CEMDEpzGIWTubQOLSzsoBaOwl

jaapjansma:
Really cool this feature! Well done Michael!

src/CRM/CivixBundle/Resources/views/Code/module.civix.php.php
```diff
@@ -446,3 +446,26 @@ function _<?php echo $mainFile ?>_civix_civicrm_alterSettingsFolders(&$metaDataF
...
+  if($k){
+    $entityTypeLines[$k] = '  '.$l;
+  }else{
+    $entityTypeLines[$k] = $l;
+  }
+}
+$entityTypes = implode("\n", $entityTypeLines);
+?>
+function _<?php echo $mainFile ?>_civix_civicrm_entityTypes(&$entityTypes) {
+  $entityTypes += <?php echo $entityTypes ?>;
```
totten: I like the idea of aggregating the list ahead of time to avoid extra file-scans. 

This should be an `array_merge()` instead of `+` -- to my reading of `CRM_Utils_Hook::entityTypes(&$entityTypes)`, there's no specific key for `$entityTypes`, so we should assume numeric keys. For numeric keys, the `+` will cause values to [trample each other](https://gist.github.com/totten/22b2d64f1a11b1da28dc4fabf7ff39e0).

michaelmcandrew: Good point - that is now updated...

totten:
@michaelmcandrew I added a bit to `tests/make-example.sh` so that it produces a working entity with the new command. Also, did a few trial runs with different mix of options (e.g. with 1 entity and with 0 entities in the extension), and it seemed to work.

Let's give this a go!

michaelmcandrew:
Tests sound good. I'm saddled up! Thanks for merging :)

PR #125, opened 2018-02-26
Bug in form template path

michaelmcandrew:
Hi again,

I am surprised this has slipped through undetected/reported. Maybe I am missing something but I think this was a bug / is a fix:

**Before:**
```
buildkit@civicrm:~/build/chatbot/sites/all/modules/civicrm/tools/extensions/civicrm-chatbot$ civix generate:form Admin_Facebook civicrm/admin/chat/facebook
...
Write /buildkit/build/chatbot/sites/all/modules/civicrm/tools/extensions/civicrm-chatbot/templates/CRM/Chatbot/Form/Admin_Facebook.tpl
```
**After**
```
buildkit@civicrm:~/build/chatbot/sites/all/modules/civicrm/tools/extensions/civicrm-chatbot$ civix generate:form Admin_Facebook civicrm/admin/chat/facebook
Write /buildkit/build/chatbot/sites/all/modules/civicrm/tools/extensions/civicrm-chatbot/templates/CRM/Chatbot/Form/Admin/Facebook.tpl
```

### DIFF
--- a/src/CRM/CivixBundle/Command/AddFormCommand.php
+++ b/src/CRM/CivixBundle/Command/AddFormCommand.php
@@ -24,6 +24,7 @@ class AddFormCommand extends AbstractAddPageCommand {
   }
 
   protected function createTplName(InputInterface $input, $ctx) {
+    $namespace = str_replace('/', '_', $ctx['namespace']);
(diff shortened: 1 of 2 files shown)


totten:
Yeah, I've ever tried to use it that way.

Note: I pushed up another commit so that`make-example.sh` includes this syntax.

PR #127, opened 2018-03-02
generate:entity - Table name should be in camel-case

michaelmcandrew:
`civix generate:entity ChatUser`

Previously made table names like `civicrm_chatuser`

Now makes table names like `civicrm_chat_user`

### DIFF
--- a/src/CRM/CivixBundle/Command/AddEntityCommand.php
+++ b/src/CRM/CivixBundle/Command/AddEntityCommand.php
@@ -12,6 +12,8 @@ use CRM\CivixBundle\Builder\Info;
 use CRM\CivixBundle\Builder\PhpData;
 use CRM\CivixBundle\Builder\Template;
 use CRM\CivixBundle\Utils\Path;
+use CRM\CivixBundle\Utils\Naming;
+
 use Exception;
 
 class AddEntityCommand extends \Symfony\Component\Console\Command\Command {
@@ -51,7 +53,7 @@ class AddEntityCommand extends \Symfony\Component\Console\Command\Command {
     }
 
     $ctx['entityNameCamel'] = ucfirst($input->getArgument('<EntityName>'));
-    $ctx['tableName'] = 'civicrm_' . strtolower($input->getArgument('<EntityName>'));
+    $ctx['tableName'] = 'civicrm_' . Naming::camelToSnake($input->getArgument('<EntityName>'));
     if (function_exists('civicrm_api_get_function_name')) {
       $ctx['apiFunctionPrefix'] = strtolower(civicrm_api_get_function_name($ctx['entityNameCamel'], '', self::API_VERSION));
     }
--- a/src/CRM/CivixBundle/Command/AddEntityCommand.php
+++ b/src/CRM/CivixBundle/Command/AddEntityCommand.php
@@ -22,9 +22,18 @@ class AddEntityCommand extends \Symfony\Component\Console\Command\Command {
   protected function configure() {
     $this
       ->setName('generate:entity')
-      ->setDescription('Add a new API/BAO/GenCode entity to a CiviCRM Module-Extension (*EXPERIMENTAL AND INCOMPLETE*)')
+      ->setDescription('Add a new API/BAO/GenCode entity to a CiviCRM Module-Extension (*EXPERIMENTAL*)')
       ->addArgument('<EntityName>', InputArgument::REQUIRED, 'The brief, unique name of the entity")')
-      ->addOption('table-name', NULL, InputOption::VALUE_OPTIONAL, 'The SQL table name. (Default: derived from entity name)');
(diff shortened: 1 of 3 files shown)


totten:
I think this policy makes sense. 👍 

FWIW, I wanted to see how `camelToSnake()` worked with various inputs, so I added a test function and pushed it to your branch. This showed a discrepancy between with how the other `Naming` functions handled snake/camel case:

```
$ phpunit4 src/CRM/CivixBundle/Utils/NamingTest.php
PHPUnit 4.8.21 by Sebastian Bergmann and contributors.

..........................................F.....

Time: 227 ms, Memory: 13.75Mb

There was 1 failure:

1) CRM\CivixBundle\Utils\NamingTest::testCamelToSnake with data set #6 ('foo-2-foo', true, 'foo_2_foo', 'Foo2Foo')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'foo_2_foo'
+'foo2_foo'

/home/myuser/src/civix/src/CRM/CivixBundle/Utils/NamingTest.php:59

FAILURES!
Tests: 48, Assertions: 27, Failures: 1.
```

Best is case probably to update `camelToSnake()` to be more consistent. But if you think there's a really good reason for it to work differently, then I suggest updating `testCamelToSnake()` to use different examples.

michaelmcandrew:
Good idea to add tests.

I think we might be hitting up against an limitation of CamelCase rather than a weakness in the function: since numbers do not have upper and lower case, we don't know if a number should start a new word.

We can
* say that digits should be lowercase (Sport4All would be misinterpreted)
* say that digits should be uppercase (Http2 would be misinterpreted)
* recommend against / prohibit digits in entity names
* not attempt to deal with the edge case. Not run tests where the input of camelToSnake contains a digit.

Thoughts?

michaelmcandrew:
aside: Just realised that I have a CamelCase surname :)

totten:
Yeah, the mechanism of mapping camel<=>snake is a bit fraught because the two spaces don't really map one-to-one.

TBH, I don't mind picking arbitrarily between the mapping behaviors for this use-case (i.e. for generating a table-name based on an entity-name). My real concerns are:

 * Someone looks as `Naming.php` and assumes it's a flexible/general-purpose set of string utilities for converting any names. It's not -- because such a thing would be problematic to define.
 * Someone is unable to generate the table name they're expecting -- because the mapping is a bit arbitrary.

I've push up a couple changes.

src/CRM/CivixBundle/Command/AddEntityCommand.php
```diff
@@ -51,7 +54,7 @@ protected function execute(InputInterface $input, OutputInterface $output) {
     }
 
     $ctx['entityNameCamel'] = ucfirst($input->getArgument('<EntityName>'));
-    $ctx['tableName'] = 'civicrm_' . strtolower($input->getArgument('<EntityName>'));
+    $ctx['tableName'] = $input->getOption('table-name') ? $input->getOption('table-name') : Naming::createTableName($input->getArgument('<EntityName>'));
```
michaelmcandrew: yeah - this makes sense

src/CRM/CivixBundle/Command/AddEntityCommand.php
```diff
@@ -21,7 +23,8 @@ protected function configure() {
     $this
       ->setName('generate:entity')
       ->setDescription('Add a new API/BAO/GenCode entity to a CiviCRM Module-Extension (*EXPERIMENTAL AND INCOMPLETE*)')
-      ->addArgument('<EntityName>', InputArgument::REQUIRED, 'The brief, unique name of the entity")');
+      ->addArgument('<EntityName>', InputArgument::REQUIRED, 'The brief, unique name of the entity")')
+      ->addOption('table-name', NULL, InputOption::VALUE_OPTIONAL, 'The SQL table name. (Default: derived from entity name)');
```
michaelmcandrew: How about we add something like: "In most cases it makes sense to let civix derive this automatically from <EntityName>. In some instances (such as when the Entity contains a number) you may prefer to explicitly set a table name." to the usage?

totten: @michaelmcandrew It's hard to fit a long explanation into the per-option descriptions; but maybe add a `setHelp()` section? Then it could show a few examples for comparison, like `civix generate:entity ChatLog` and `civix generate:entity SMSLog --table-name=civicrm_sms_log`.

michaelmcandrew:
I realised/remembered that consecutive capitals are also problematic.

CiviIAmAnEdgeCase should turn into civi_i_am_an_edge_case

but CiviTWA should probably be civi_twa.

I added some usage to account for both.

michaelmcandrew:
I also removed the incomplete from `generate:entity` as it is IMO complete now albeit still experimental

PR #128, opened 2018-03-07
(Un)set version in entity XML templates

michaelmcandrew:
I thought this might be a reasonable way of getting an appropriate version for the xml file.

### DIFF
--- a/src/CRM/CivixBundle/Resources/views/Code/entity-schema.xml.php
+++ b/src/CRM/CivixBundle/Resources/views/Code/entity-schema.xml.php
@@ -1,6 +1,6 @@
 <?php
 echo '<?xml version="1.0" encoding="iso-8859-1" ?>'."\n";
-$fixmeVersion = '4.4';
+$version = civicrm_api3('Extension', 'getvalue', ['key' => $mainFile, 'return' => 'version']);
 ?>
 
 <table>
@@ -8,7 +8,7 @@ $fixmeVersion = '4.4';
   <class><?php echo $entityNameCamel ?></class>
   <name><?php echo $tableName ?></name>
   <comment>FIXME</comment>
-  <add><?php echo $fixmeVersion ?></add>
+  <add><?php echo $version ?></add>
   <log>true</log>
 
   <field>
@@ -16,7 +16,7 @@ $fixmeVersion = '4.4';
     <type>int unsigned</type>
     <required>true</required>
     <comment>Unique <?php echo $entityNameCamel ?> ID</comment>
-    <add><?php echo $fixmeVersion ?></add>
+    <add><?php echo $version ?></add>
   </field>
   <primaryKey>
     <name>id</name>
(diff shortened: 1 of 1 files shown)


src/CRM/CivixBundle/Resources/views/Code/entity-schema.xml.php
```diff
@@ -1,22 +1,22 @@
 <?php
 echo '<?xml version="1.0" encoding="iso-8859-1" ?>'."\n";
-$fixmeVersion = '4.4';
+$version = civicrm_api3('Extension', 'getvalue', ['key' => $mainFile, 'return' => 'version']);
```
totten: So I'm really unsure about this one. Random reactions:

* Is this just written as a WIP/gist, or does it actually work?
* The `<add>` / `<drop>` have traditionally been versions of `civicrm-core`.
* Does any process actually use this information right now? For example, scanning metadata? If so, are they more likely to break if they (a) continue to see core version#s or (b) see extension version#s?
* In the metadata for the entity, it would be useful to have some indication of origin.
* I usually try to avoid a booting/needing a running a Civi if I can avoid it, but it's not critical. I guess that `generate:entity` already. boots up to use`_civicrm_api_get_entity_name_from_camel()`.

CC @colemanw @eileenmcnaughton

michaelmcandrew:
maybe i am getting a bit trigger happy. it is not massively thought through, but I didn't like having the random number 4.4 in my extension xml files and I figured that the actual version pulled from the extension's xml file is better than 4.4 and "It works for me".

> Does any process actually use this information right now? For example, scanning metadata? If so, are they more likely to break if they (a) continue to see core version#s or (b) see extension version#s?

A few days ago, some process that I was running was complaining when it wasn't present but I cannot reproduce the complaint now. It was checking the version was < or > than something else (I can't recall what :crying_cat_face:)

> I usually try to avoid a booting/needing a running a Civi if I can avoid it, but it's not critical. I guess that generate:entity already. boots up to use_civicrm_api_get_entity_name_from_camel().

Oh, OK. I had presumed that civi was booted in a few of the commands so didn't think it was a big deal. We could parse the xml file instead :thinking:

totten:
Agree the `<add>4.4</add>` feels pretty random (and serves no particular purpose). But IMHO, if we set this to the extension version, then that represents a semantic change to the XML schema. I *suspect* it's safe, and the `generate:entity` is presented as "Experimental", but I'd prefer to get a quick reaction from someone else who's worked with that metadata.

totten:
Related: https://chat.civicrm.org/civicrm/pl/n95mb7a44pfrmno95jpyop3y7w

src/CRM/CivixBundle/Resources/views/Code/entity-schema.xml.php
eileenmcnaughton: I don't think it actually does anything currently....

eileenmcnaughton:
I would think remove makes sense....

michaelmcandrew:
Hey there, agree about removal.

@eileenmcnaughton - I did try and remove them when using the old method (using the core gencode process) but it complained they were not there so I presumed they would be needed now.

@totten - I just pushed another commit to this branch that removes them. I tested this with `generate:entity-boilerplate` and there was no change in the generated sql/auto_{install,uninstall).sql so I think this is good to go.

michaelmcandrew:
wannamerge?

totten:
Sure!

🤘

PR #129, opened 2018-03-15
Remove select function override with more specific Select function

eileenmcnaughton:
We don't really encourage people to override select unless they are adding value. There are further trim downs this report could do with to reduce boilerplate code that I may get to later.

Note the parent function is better than the override function so we should use that


src/CRM/CivixBundle/Resources/views/Code/report.php.php
```diff
@@ -40,10 +40,6 @@ class <?php echo $reportClassName ?> extends CRM_Report_Form {
             'title' => E::ts('Last Name'),
             'no_repeat' => TRUE,
           ),
-          'id' => array(
```
eileenmcnaughton: duplicate field

totten:
Closing this because it seems to overlap with the newer/broader #132 (and is now conflicted because #132 was merged).

PR #130, opened 2018-03-21
Fix non-sensical variable setting

eileenmcnaughton:
The lines moved are meaningless unless moved per this PR (arguably they should be removed or the function that uses these determines the version, but if they are there they should be right)


eileenmcnaughton:
@ejegg....

ejegg:
+1 looks good to me!

totten:
Yup, that makes much more sense. 👍 

FWIW, I still think it's preferable for any `*.mgd.php` files to explicitly set the `version`, so the various code-generators should continue doing that. But defaulting to `version=>3` is reasonable and it's in the code, so it should probably work.

PR #132, opened 2018-04-12
Update report php template

eileenmcnaughton:
Improve template to reflect recommended practice - ie. don't override main select, where order by functions but use specific field overrides instead. Allow built in group by & order by to work by default.


seamuslee001:
I trust Eileen here and this seems to make sense

totten:
Thanks for the nudge @seamuslee001 and for the update @eileenmcnaughton. I'm also inclined to take Eileen's word on this. 👍

PR #137, opened 2018-05-09
More helpful output for generate:entity command

michaelmcandrew:
The following output more clearly directs the user as to appropriate next steps.


totten:
Good idea, @michaelmcandrew.

PR #138, opened 2018-05-23
order mgd files before including

michaelmcandrew:
Order is important when managing entities: one must create dependencies before creating dependent entities. Custum groups need to be created before custom fields, option groups before option values, etc.

One way of doing this is to ensure that dependencies appropriate order in the same file. This breaks down if dependencies are split across files.

Alphabetically sorting the managed files before including them allows one to define managed entities like this
```
managed
├── 01-ContactType.mgd.php
├── 05-CustomGroup.mgd.php
└── 06-CustomField.mgd.php
```

Aside: a more sophisticated `CRM_Core_ManagedEntities` might be cleverer about trying to 'satisfy dependencies' before giving up ( [something like this](https://gist.github.com/michaelmcandrew/21754058092bfd63504490f59ae19944)) but in the mean time, this is a decent approach, IMO.

### DIFF
--- a/src/CRM/CivixBundle/Resources/views/Code/module.civix.php.php
+++ b/src/CRM/CivixBundle/Resources/views/Code/module.civix.php.php
@@ -266,6 +266,7 @@ function _<?php echo $mainFile ?>_civix_find_files($dir, $pattern) {
  */
 function _<?php echo $mainFile ?>_civix_civicrm_managed(&$entities) {
   $mgdFiles = _<?php echo $mainFile ?>_civix_find_files(__DIR__, '*.mgd.php');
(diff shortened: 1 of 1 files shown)


totten:
@michaelmcandrew I'm not sure why I didn't notice this one earlier. Agree with this.

The organic ordering produced by a file-system can be totally bonkers -- which would mean that the same codebase could execute with different orders in different deployments. You're right that something more sophisticated in core could handle the sequencing better... but in absence of that, this should at least ensure that the behavior is consistent. :+1:

michaelmcandrew:
cool, thanks for merging

Issue #139, opened 2018-06-13
_hook_civix_civicrm_alterSettingsFolders only effective on first invocation


michaelmcandrew:
_hook_civix_civicrm_alterSettingsFolders checks to see if it has already been called and if so, it returns early without adding directories to $metaDataFolders

This logic breaks down for any pathways where it gets called twice (I've come across two of these - there might be others).

A simple solution is to remove the 'is configured' check.

```php
  static $configured = FALSE;
  if ($configured) {
    return;
  }
  $configured = TRUE;
```

A less simple solution (if we are worried about the extra call to is_dir() call would be

```php
  static $configured = FALSE;
  static $settingsDir = NULL;

  if (!$configured) {
    $settingsDir = __DIR__ . DIRECTORY_SEPARATOR . 'settings';
    if (!is_dir($settingsDir)) {
      $settingsDir = NULL;
    }
    $configured = TRUE;
  }
  if ($settingsDir && !in_array($settingsDir, $metaDataFolders)) {
    $metaDataFolders[] = $settingsDir;
  }
```
Both the above solutions leave extensions that currently depend on this functionality liable to break.

An alternative approach would be to patch core so that CRM_Utils_Hook::alterSettingsFolders($metaDataFolders); is only ever called once and we cache the result, but not sure how feasible that is.

michaelmcandrew:
I'm thinking that we might want to include an upgrade message with the next version of civix to address this (suggest people edit their extension.civix.php files)

totten:
@michaelmcandrew good analysis.

(For anyone reading newly, the full text of the template being discussed is...)

https://github.com/totten/civix/blob/008e9aa61c0040f2edb5392c0bd1d51a15a9de8c/src/CRM/CivixBundle/Resources/views/Code/module.civix.php.php#L439-L450

TBH, I don't really know why it has a `static $configured` guard. Tracing the records, it wasn't part of a bugfix in civix -- it was apparently copied-pasted from the original exemplar in #29. @eileenmcnaughton might have a better idea.

My suspicion -- it borrowed from `_hook_civix_civicrm_config` which uses a `static $configured` guard as a very fast way to prevent duplicate changes in global state. However, IMHO, there are few differences between these:

1. `h_c_alterSettingsFolder` has another guard (`is_dir(...) && !in_array(...)`) which looks more substantive and correct.
[...]

PR #140, opened 2018-06-18
remove the static $configured guard from h_c_alterSettingsFolder

michaelmcandrew:
see https://github.com/totten/civix/issues/139

I don't think a upgrade notice is that necessary. I was forgetting that that file is auto-generated and easy to refresh.

That said, if lots of people start getting bitten by this, we might want to reconsider.


eileenmcnaughton:
I agree with this change but don't have merge rights. I'm not too worried about the upgrade notice - it's not new

eileenmcnaughton:
Oh I see there is one already open but I couldn't merge - @totten

PR #141, opened 2018-07-09
Bump default version to 5.0

eileenmcnaughton:
arguably this should be 5.3 as earlier versions will be insecure


seamuslee001:
I think this is ok welcome the discussion around whether its 5.0 or 5.3

eileenmcnaughton:
I would go 5.3 - but either is an improvement.

My preferred would be the current civi version - ie. if you are writing against 5.6 you don't want to imply it will work on 5.4 - but I think it's not important to get the best outcome - just better than 4.7

totten:
Agree with all above. It's patch-welcome to initialize the `<ver>` programmatically (to match `CRM_Util_System::version()` as long as the system has bootstrapped). In the mean time, defaulting to `5.0` seems better than `4.7.`

PR #142, opened 2018-07-30
Add use of use \Civi\Test\Api3TestTrait to make testwriting easier

eileenmcnaughton:


seamuslee001:
This looks good to me @totten

src/CRM/CivixBundle/Resources/views/Code/test-api.php.php
```diff
@@ -12,6 +12,7 @@
  * @group headless
  */
 class <?php echo $testClassName ?> extends \PHPUnit_Framework_TestCase implements HeadlessInterface, HookInterface, TransactionalInterface {
+  use \Civi\Test\Api3TestTrait;
```
seamuslee001: @eileenmcnaughton should we include the DocTrait as well?

eileenmcnaughton: @seamuslee001 possibly - but this is like from July & I don't even remember it now so I don't want to scope creep it at the moment

totten:
Thanks, @eileenmcnaughton  @seamuslee001. Agree this makes sense.

PR #146, opened 2019-02-09
Use helper function for angular module requirements

colemanw:
Before
----
Requirements declared in js

After
------
Requirements declared in php


totten:
Tested and looks good.

PR #148, opened 2019-03-06
Add api4 file when creating an entity

colemanw:
This adds an api4 file to expose new civix entities to api v4.

### DIFF
--- a/src/CRM/CivixBundle/Builder/Info.php
+++ b/src/CRM/CivixBundle/Builder/Info.php
@@ -88,4 +88,13 @@ class Info extends XML {
     return (string) $attrs['type'];
   }
 
+  /**
+   * Get the user-friendly name of the extension.
+   *
+   * @return string
+   */
+  public function getExtensionName() {
+    return empty($this->xml->name) ? 'FIXME' : $this->xml->name;
+  }
+
 }
--- a/src/CRM/CivixBundle/Command/AddEntityCommand.php
+++ b/src/CRM/CivixBundle/Command/AddEntityCommand.php
@@ -74,16 +74,19 @@ explicity.');
       throw new Exception("Failed to determine proper API function name. Perhaps the API internals have changed?");
     }
     $ctx['apiFile'] = $basedir->string('api', 'v3', $ctx['entityNameCamel'] . '.php');
+    $ctx['api4File'] = $basedir->string('Civi', 'Api4', $ctx['entityNameCamel'] . '.php');
     $ctx['daoClassName'] = strtr($ctx['namespace'], '/', '_') . '_DAO_' . $input->getArgument('<EntityName>');
     $ctx['daoClassFile'] = $basedir->string(strtr($ctx['daoClassName'], '_', '/') . '.php');
     $ctx['baoClassName'] = strtr($ctx['namespace'], '/', '_') . '_BAO_' . $input->getArgument('<EntityName>');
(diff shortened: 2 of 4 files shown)


colemanw:
@totten this is ready to merge

totten:
@colemanw @seamuslee001  OK, so when I first skimmed this, I felt a little confused/surprised at having a DX generating both; but looking again, it's making some more sense. The tension:

* On one hand, v3/v4 are different things -- and there are foreseeable occasions where one supports v3 or v4 or both. Any QA or behavioral tweaks have to be different for the v3 and v4 variants. (You probably shouldn't generate an API which you're not going to maintain.) So from that perspective, I expect the v3+v4 generators to be independently addressable. 
    * (For example -- fast forward 2 years, you'd probably want it to generate APIv4 -- and not APIv3. Today, there's probably a cohort who expect v3 and will be surprised to see v4.)
* On the other hand, there's gonna be a long period of v3/v4 coexistence/transition. Generating both by default makes it easier to get people to start using v4. That's a legit reason to generate both at the start of the transition period.

(Still haven't formed an opinion on best approach... but feel better about understanding...)

seamuslee001:
I feel that the best outcome at the moment is we generate each, i think when we feel that apiv3 enters an end of life for new APIs / api4 matures then we can look to remove the generation of apiv3, We should also look at a new action like generate:api4 to generate just an APIv4 class

eileenmcnaughton:
Can  we merge this? I can't see  any blockers  since  it's not removing anything (at  this stage)

totten:
My expectation is the default API generated for an entity is "scaffolding" but not secure/correct by default - by creating both, we're putting an onus on the user to recognize this and then choose whether to (a) delete the v3 variant, (b) delete the v4 variant, or (c) maintain both variants. The problem is that - at the time when they use civix - dev's often aren't in a great position to intuit which files are good/bad/useful/why.

But I take this PR-discussion to mean generally that the API maintainers want v4 by default. That's cool.

I'm leaning toward this DX:

* `civix generate:entity` accepts an option `--api-version`
* `--api-version` is accepts a list (eg `3` or `4` or `3,4`)
* `--api-version` defaults to `4`
* When you run `civix generate:entity`, it shows messages to (a) specifically indicate which API versions are/are-not generated and (b) encourage user to set `--api-version` if they care.

eileenmcnaughton:
I guess that makes sense - I would probably go with  both transitionally while people get familiar with apiv4 but that's just my preference based on trying to start that transition myself

eileenmcnaughton:
Hmm looks like today will not be the day I create  my own apiv4 entity - I tried creating one without the  benefit of this  - because I  hadn't found  it  - and  I seem to be having autoloader issues  & have hit the 'I've spent enough time trying to do apiv4 instead of apiv3' point where I will fail back to apiv3 & move on.

I would note this is only part of what we need - we also need 

```
 civix generate:api myentity import
```

Anyway - my vote is to merge this as  is because it's been sitting around since March in a state that would have provided benefit had it been merged earlier - but which is arguably less good than it might be

colemanw:
I'd also vote for merging this now & refining it based on Tim's suggestion later.
Like api3, new api4 entities inherit the default ultra-strict permission of "administer CiviCRM" so adding this isn't going to open up any security holes.

totten:
I've pushed up the changes to accept `--api-version=X,Y,Z`. (Shorthand: `-A3` or `-A4` or `-A3,4`). Merging.

totten:
Published civix v19.08.0

PR #150, opened 2019-03-19
#139 Improve alterSettingsFolders hook.

eileenmcnaughton:
The use of static caching in this function is unhelpful for tests & we get the same
benefit by moving the in_array check to before the
is_dir check - meaning we won't do a file check unless it is not present

(Just found this patch locally & realised it must have been written when agreed in #139 but never opened - @totten I can't merge to this repo)


totten:
Yeah, it does look pretty reasonable and was generally agreed in #139.

PR #153, opened 2019-06-04
updates for Windows

demeritcowboy:
Semicolon doesn't separate commands on windows but && should work everywhere.

As an aside it's been a while since I updated this fork, but at least I'm consistent ... he he

58841559-69090480-8638-11e9-92be-522f7aed7f95


demeritcowboy:
Closing since while the change does improve portability there's something else still going on where it doesn't work as-is on windows.

PR #154, opened 2019-06-06
Update angular page template to use controllerAs - to reflect best practice

eileenmcnaughton:
Top item when you look up angular best practices is to use controllerAs syntax - effectively namespacing.

This is also a thing if we ever want to migrate to angular.

Not sure the impact on the prettiness of the example but we should comply somehow with best practice


colemanw:
From what I've read, standalone controllers themselves are deprecated in favor of components. That's annoying with Angular 1 + routing because as far as I can tell, routes *require* a standalone controller.

At least adding `controllerAs`, as you've done here, is a step in the right direction. Although I'd lean toward the default value used by components for consistency, which is `"$ctrl"`.

colemanw:
@totten I recommend closing this in favor of #187

PR #155, opened 2019-06-12
Update unit test class for phpunit 6.x support

eileenmcnaughton:


totten:
Agree with this.

On the communications front -- I sent PSA about this a couple weeks ago to `civicrm-dev` (2019-06-07) and added a shpiel in `UPGRADE.md`.

PR #156, opened 2019-06-14
generate:theme - Add subcommand for a skeletal theme definition

totten:


eileenmcnaughton:
@totten merge merge merge merge

totten:
Rebased and updated stale bits in UPGRADE.md

On policy/concept level, agree with merging. But it has been a year since the code was written - merits some kind of `r-run`. Happy to merge if someone else has tried it out recently. If not, it may still be a couple days before I give it a proper whirl.

seamuslee001:
@totten

a couple of things

1. when i ran civix generate::theme - the only files that were created / modifled were civicrm.css bootstrap.css in a css/ folder and shortname.theme.php file, there were no changes to the civix file at all.

2. when i implemented the hook manually it is not clear that a) the key of the $themes array needs to match the name (maybe this is taken care of in civix) 2. the data in the .theme file is missing the ext key but i presume that might get added through the civix set of function a la civicrm_managed

However once i got my hook manually right (without civix) then it all worked as expected

totten:
Yeah, filling-in hook stubs retroactively is a bit annoying. This *should* follow the same flow as when we've had to add other hooks, i.e.

1. If you make a new extension, then the code-generator "just works" -- because the new extension is generated with the new hook-stubs.
2. If you update an existing extension, then it does not update `myext.php`. You need to do a few things mentioned in https://github.com/totten/civix/blob/bb15c3752d47d193dead960e8bf762fea7075931/UPGRADE.md :
    1. "General Tasks" => i.e. `civix generate:module` to regenerate the stubs in `myext.civix.php`
    2. "Special Tasks: Upgrade to v19.06.2+: hook_civicrm_themes" => i.e. manually add a stub in `myext.php`

Of course, it's generally annoying that one has to do these things, and it'd be great to have a better workflow for filling in new hooks, but IMHO that's a distinct problem that doesn't need to block this one.

The Q -- if you have a new extension and follow the `UPGRADE.md`, does it give you working stubs?

seamuslee001:
@totten yep confirmed after running `civix generate::module` in my extension and then putting in the boiler plate code it all works and confirmed it works correctly out of the box doing

```
civix generate:module
civix genearte:theme
```
produces the desired outcome

totten:
Brilliant. Thanks @seamuslee001 .

PR #162, opened 2019-10-08
generate:module creates skeletal mkdocs content

ejegg:
Just generates mkdocs.yml and docs/index.md with a couple of Template
builders in the main InitCommand for now. Alternatively, one could create
a Docs builder and/or command.

Fixes #161

PR #163, opened 2019-10-08
Update mkdocs.yml template for mkdocs 2.5+

ejegg:
The markdown_extensions syntax has changed, but the current publishing infrastructure doesn't support the new stuff yet. Splitting this into a separate PR

### DIFF
--- a/src/CRM/CivixBundle/Resources/views/Code/mkdocs.yml.php
+++ b/src/CRM/CivixBundle/Resources/views/Code/mkdocs.yml.php
@@ -1,0 +1,19 @@
+site_name: <?php echo "$fullName\n"; ?>
+repo_url: https://lab.civicrm.org/extensions/FIXME
+theme: material
+
+pages:
+- Home: index.md
+
+markdown_extensions:
+  - attr_list
+  - admonition
+  - def_list
+  - codehilite(guess_lang=false)
+  - toc(permalink=true)
+  - pymdownx.superfences
+  - pymdownx.inlinehilite
+  - pymdownx.tilde
+  - pymdownx.betterem
+  - pymdownx.mark
+  - pymdownx.magiclink
--- a/src/CRM/CivixBundle/Resources/views/Code/mkdocs.yml.php
+++ b/src/CRM/CivixBundle/Resources/views/Code/mkdocs.yml.php
@@ -9,8 +9,10 @@ markdown_extensions:
   - attr_list
   - admonition
   - def_list
-  - codehilite(guess_lang=false)
-  - toc(permalink=true)
+  - codehilite
+      guess_lang: false
+  - toc:
+      permalink: true
   - pymdownx.superfences
   - pymdownx.inlinehilite
   - pymdownx.tilde
--- a/src/CRM/CivixBundle/Command/InitCommand.php
+++ b/src/CRM/CivixBundle/Command/InitCommand.php
@@ -108,6 +108,7 @@ class InitCommand extends AbstractCommand {
     $basedir = new Path($ctx['basedir']);
     $ext->builders['dirs'] = new Dirs([
       $basedir->string('build'),
+      $basedir->string('docs'),
       $basedir->string('templates'),
       $basedir->string('xml'),
       $basedir->string('images'),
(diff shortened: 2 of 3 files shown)


src/CRM/CivixBundle/Resources/views/Code/mkdocs.yml.php
```diff
@@ -0,0 +1,21 @@
+site_name: <?php echo "$fullName\n"; ?>
+repo_url: https://lab.civicrm.org/extensions/FIXME
+theme: material
+
+pages:
+- Home: index.md
```
totten: @ejegg 

Would it make sense to point this at the README.md file? There's a lot of overlap between the `index.md` and `README.md` templates (wrt general overview and requirements).

https://github.com/totten/civix/blob/master/src/CRM/CivixBundle/Resources/views/Code/readme.md.php

There are a few differences wrt to install-instructions, usage, known issues, and future plans - however, I think these differences are more happenstance than essence.  Those topics all seem equally plausible as "quick intro material".

src/CRM/CivixBundle/Resources/views/Code/mkdocs.yml.php
```diff
@@ -0,0 +1,21 @@
+site_name: <?php echo "$fullName\n"; ?>
```
totten: Maybe leave a trail pointing to some information about what how to use mkdocs? e.g.

```suggestion
## Use mkdocs to generate a manual for this extension. For more information about mkdocs,
## see https://docs.civicrm.org/dev/en/latest/documentation/

site_name: <?php echo "$fullName\n"; ?>
```

src/CRM/CivixBundle/Resources/views/Code/mkdocs.yml.php
ejegg: It might make sense! At this point, I'm just trying to be consistent with the instructions at https://docs.civicrm.org/dev/en/latest/documentation/extensions/ .
In the README section on that page, it suggests that you might want to just make a symlink from README.md to docs/index.md. So I guess I'd want to leave this yml file as-is in any case, just start with a symlink.

src/CRM/CivixBundle/Resources/views/Code/mkdocs.yml.php
ejegg: Good call. I've updated it in PR #162, which just has the first commit introducing these files with the old mkdocs.yml extensions syntax. I'll rebase this one once that one gets merged.

src/CRM/CivixBundle/Resources/views/Code/mkdocs.yml.php
MikeyMJCO: Without a file called "index.md" you won't get an autoloaded page when hitting the repo. There needs to be at least one file called "index" at least at the moment.

ejegg: Thanks @MikeyMJCO! over in PR #162 (the version that should be mergeable before the docs infrastructure upgrade) I left index.md in place and just added the explanatory comment suggested below.

MikeyMJCO:
This can be merged - current docs infra supports all of this.

eileenmcnaughton:
@totten see ^^ we should merge

ejegg:
OK, this is rebased + title changed to remove DNM warning

src/CRM/CivixBundle/Resources/views/Code/mkdocs.yml.php
```diff
@@ -0,0 +1,21 @@
...
+repo_url: https://lab.civicrm.org/extensions/FIXME
+theme: material
+
+pages:
```
MikeyMJCO, line 5:
```suggestion
pages:
```

The pages key is deprecated. It won't work. This should use `nav:`

MikeyMJCO (commented):
This also needs to add the theme and category to the template.

MikeyMJCO:
See the example here: https://docs.civicrm.org/dev/en/latest/extensions/documentation/#config

MikeyMJCO:
The file should have the `site_name`, `site_description`, `theme:`, `nav:`, `markdown_extensions:` and `plugins:` keys.

ejegg:
Oh hey @MikeyMJCO, since PR #162 with the older version never got merged, there's no need to do this in two commit. I've just updated #162 with the new syntax and the extra keys from that example. Let's merge that one instead of this one.

PR #164, opened 2019-10-15
Replace NewEntity and NewAction with the names of that being generated

artfulrobot:
This is the little change I promised to offer to make!

I think the diff is small enough to explain more clearly than I could in words


eileenmcnaughton:
@totten I tested this & it works

totten:
Yup, it also worked for me.

totten:
Thanks @artfulrobot !

PR #165, opened 2019-10-22
Remove extra function parameter when inserting submenus

ejegg:
_bla_civix_insert_navigation_menu only takes 3 params; don't call it with 4


eileenmcnaughton:
@totten I've confirmed this - can you merge it?

eileenmcnaughton:
Tested too

totten:
Oh, good call. :+1: Thanks @ejegg

PR #166, opened 2019-10-22
Simplify install steps and link to docs

mattwire:
Most of the readme is taken up by docs for installing extension in various different ways.  The developer specific ones are probably less useful because each dev does things slightly differently (I don't follow the instructions here).  Better to point to our CiviCRM documentation which can be dynamically updated for installing extensions.


eileenmcnaughton:
I'm +1 on this

eileenmcnaughton:
OK let's track this here - https://github.com/civicrm/civicrm-dev-docs/issues/713 - 
(we can close this PR until that is resolved)

PR #167, opened 2019-10-22
Style cleanup

eileenmcnaughton:
- fix array syntax (array() => [])
- brings comment block formatting closer to drupal standard (higher level than we enforce)
- adds a missing return tag
- replaces a CRM_Core_Fatal with throwing an exception - we've been doing this pretty aggressively in core so probably time to do here too. Bonus, it's in a function that says 'Note: This hook only runs in CiviCRM 4.4+.' and should probably be removed anyway by now


totten:
Yup, checks out for me.

PR #168, opened 2019-10-23
Api file cleanup

eileenmcnaughton:
@totten  a little more cleanup
- array syntax
- comment block tidying
- we don't really recommend error codes so removed that
- we don't return civicrm_api3_create_error so removed that too

### DIFF
--- a/src/CRM/CivixBundle/Resources/views/Code/api.php.php
+++ b/src/CRM/CivixBundle/Resources/views/Code/api.php.php
@@ -9,7 +9,7 @@ use <?php echo $_namespace ?>_ExtensionUtil as E;
  * This is used for documentation and validation.
  *
  * @param array $spec description of fields supported by this API call
- * @return void
+ *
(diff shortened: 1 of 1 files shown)


totten:
Agree with the syntax cleanups and that exceptions are generally better habit. Tested, and those bits are working. :+1:

Regarding the error-codes... I think [API_Exception::__construct(...$error_code...)](https://github.com/civicrm/civicrm-core/blob/5.18/api/Exception.php#L30) is confusing: the docblock recommends code-names/strings; the parent class requires ints; the resulting API-results magically work with either.

I actually agree with the docblock - it's good practice to have a symbolic name for each error-condition so that consumers can be *smart* about error-handling. (*Most class-models for `Exception` address this with subclassing; but that's counter-factual for APIv3. The `error_code` is what APIv3 has.*) The practice may not be used a lot in APIv3, but I think that's because *people don't think about error-handling enough*. It's still good+easy to return an error code.

Updated, retested, and merging.

PR #169, opened 2019-10-23
Fix up entity api template

eileenmcnaughton:
@totten and another one 
- comment blocks closer to full drupal spec (we don't enforce it all but for templates....)
- pass the entity name to create & get. This is pretty important as you MUST do this if you want to enable custom data for your api (you also need to add an option value). This removes a trap for young players


totten:
Yup, seems good. :+1: Cheers @eileenmcnaughton

PR #170, opened 2019-11-13
Builder\Info - Generate '<classloader>/<psr4>' to support APIv4

totten:
APIv4 looks for classes in the `Civi\Api4` namespace and `Civi/Api4` folder.  To support generation of APIv4 code, the `info.xml` should have a corresponding classloader:

```xml
  <classloader>
    <psr4 prefix="Civi\" path="Civi" />
  </classloader>
```

This revision:

* Includes the default on new modules
* Adds a note for people updating existing modules

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -41,6 +41,18 @@ The steps for upgrading the `Upgrader` are as follows:
 
 ## Special Tasks
 
+### Upgrade to v19.11.0+: APIv4 and PSR-4
+
+APIv4 looks for classes in the `Civi\Api4` namespace and `Civi/Api4` folder. 
+To support generation of APIv4 code, the `info.xml` should have a
+corresponding classloader:
+
+```xml
+  <classloader>
+    <psr4 prefix="Civi\" path="Civi" />
+  </classloader>
+```
+
 ### Upgrade to v19.06.2+: PHPUnit (Optional; #155)
 
 The templates for PHPUnit tests have been updated to match a major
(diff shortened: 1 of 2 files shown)


eileenmcnaughton:
cool

totten:
Note: I did a little `r-run` by creating an extension with an APIv4 entity (`generate:entity Foo... generate:entity-boilerplate... generate:upgrader`). With this patch, I was able to see the generated entity:

```
[bknix-min:~/.../modules/civicrm/ext/foo] cv api4 Entity.get +w name=Foo
{
    "0": {
        "name": "Foo",
        "description": "Foo entity.",
        "comment": "Provided by the FIXME extension."
    }
}
```

eileenmcnaughton:
Nice - are the extraneous spaces still present in the _civix class at the ends of the lines in the entityType hook?

totten:
Yeah, this didn't touch that. Honestly, `var_export()` won't give pretty data. The two options are:

1. Add a directive to `*.civix.php` so that phpcs just ignores it.
2. Write a replacement for `var_export()` which can be used in `Module.php` and `PhpData.php`. Something along the lines of this pseudocode:

```php
/**
 * @var mixed $var  The data to export
 * @var string $indent  Prefix indicating level of indentation for formatted output
 */
function civiStyleExport($var, $indent = '') {
  if (is_scalar($var)) return var_export($var, 1);
  elseif (is_array($var)) {
    return $indent . '[' ... recursively export each item with suitable indentation  ... ']';
  }
  else {
    throw new \RuntimeException(sprintf('%s: Unsupported data type %s',
      __FUNCTION__, gettype($var)));
  }
}
```

It's patch-welcome either way.

PR #171, opened 2019-12-11
Convert deprecated function CRM_Core_BAO_Setting::getItem

demeritcowboy:
I don't think this line gets called much, but when doing a grep every extension that has an upgrader shows up. I haven't tested this - I guess it comes up if you try to update a really old extension.


totten:
Yeah, this one is a little tricky to test b/c it only applies when migrating older metadata (ie migrate the setting `{$extension}:version` to the SQL column `civicrm_extension.schema_version`).

But I did test it, and it seems good. :+1:

demeritcowboy:
Thanks!

PR #173, opened 2020-02-17
Fix #172: SQL files not written by generate:entity-boilerplate

### LINKED ISSUE #172: generate:entity-boilerplate no longer creates SQL files
In [this line](https://github.com/totten/civix/blob/v19.11.0/src/CRM/CivixBundle/Command/AddEntityBoilerplateCommand.php#L108) there's a call to a function that generates and returns SQL.

But that SQL is not saved anywhere.

It reports that it successfully wrote sql files, but in fact it did no such thing.

artfulrobot:
This fixes the issue by calling `file_put_contents` instead of... well... *not* calling `file_put_contents`.

### DIFF
--- a/src/CRM/CivixBundle/Command/AddEntityBoilerplateCommand.php
+++ b/src/CRM/CivixBundle/Command/AddEntityBoilerplateCommand.php
@@ -25,7 +25,6 @@ class AddEntityBoilerplateCommand extends \Symfony\Component\Console\Command\Com
       );
   }
 
-
   /**
    * Note: this function replicates a fair amount of the functionality of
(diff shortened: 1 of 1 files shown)


totten:
Good catch @artfulrobot!

FWIW, this function signature [changed in 5.23.alpha1](https://github.com/civicrm/civicrm-core/pull/16405/commits/50b8ea3e511f1a35c37084133adf8c14c185baa4). Consequently, a straight patch for 5.23 compat (like 2884559) would break compat w/5.22 and earlier. This would mean that the civix and civicrm-core versions need to match exactly, which can be a bit unforgiving in transitional period.

I updated the patch to support, then ran on 5.22 and 5.23, and confirmed that it produced the same SQL content in both cases.

PR #176, opened 2020-05-28
Updated Symfony components

kewljuice:
Updated Symfony components so requirements for composer are met.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,10 +6,10 @@
     "require": {
         "php": ">=5.3.3",
         "civicrm/cv": "~0.2",
-        "symfony/console": "~2.8",
-        "symfony/process": "~2.8",
-        "symfony/templating": "~2.8",
-        "symfony/var-dumper": "~2.8",
+        "symfony/console": "~2.8|^3|^4",
+        "symfony/process": "~2.8|^3|^4",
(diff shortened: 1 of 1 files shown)


totten:
Merging.

Just to verbalize a few things:

* Symfony libraries are usually pretty good about upgrades.
* For `civix`, the normal test that I do is to run `tests/make-example.sh` and compare outcomes for a bunch of commands (`diff -ru`). On php71, I tried a `composer update` (yielding mostly Symfony 3.x stuff) and then `tests/make-example.sh`. It gave consistent output. :+1:
* But there can be [weird things in Symfony upgrades](https://github.com/civicrm/cv/pull/71). And I wish that `tests/make-example.sh` had some more nuance to it.
* The patch only updates `composer.json`; it leaves `composer.lock` alone. This means that the PHAR and standalone builds would still use the same version -- but, if you pulled in `civicrm/civix` as a dependency, then I suppose it would reconcile the versions. (I haven't tried that in a style in a long time, but if it helps - that's cool.) But sometimes it's handy to relax `composer.json` while keeping `composer.lock`. (This has been done a few times for `civicrm-core`.)
* To make the PHAR distribution more robust, I've opened #180 so that `civix` libraries (regardless of version) won't conflict with the main Civi+CMS build.

PR #180, opened 2020-06-25
Use PhpScoper to prevent library conflicts

totten:
A few `civix` subcommands involve bootstrapping Civi. This creates potential for version-conflicts in common libraries like `symfony/process`. Using `PhpScoper` means that any libraries bundled into `civix.phar` will be renamed to avoid conflicts.

See also: https://github.com/humbug/php-scoper

### DIFF
--- a/box.json
+++ b/box.json
@@ -36,6 +36,7 @@
     ],
     "git-version": "package_version",
     "main": "bin/civix",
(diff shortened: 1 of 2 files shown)


totten:
The scoping policy is a slightly modified version of `cv`'s scoping policy.

Did a run through with `tests/make-example.sh` and it looks like it produces the same output.

PR #186, opened 2020-07-09
NFC - Style fixes

colemanw:
Fixes some style issues caught by jenkins and some other ones I noticed in upgrader-base.php


totten:
Tried it out on the `make-example.sh` script - looks good. `CRM/Foo/Upgrader/Base.php` now passes `civilint`.

PR #187, opened 2020-07-09
Use controllerAs for angular best-practice

colemanw:
This is a reviewers cut of #154 - I've tested it out and confirm the new code works.


colemanw:
@totten I don't have merge rights or I would have merged this.

totten:
Yup, `r-code` and `r-run` are fine.

Fortunately, I'm pretty sure this change only affects the generated client-side page - i.e. there's no shared resource or implicit dependency with the other generators. So I think that gives us a free hand to revise this template. :+1:

PR #188, opened 2020-07-09
Update PHP. Use symfony varExporter for prettier output

colemanw:
Before
--------
Generated php files such as `*.ang.php` would not look so great, with bad indentation and old array syntax.

After
------
Should look much better.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,13 +4,14 @@
     "homepage": "https://github.com/totten/civix",
     "license": "AGPL-3.0",
     "require": {
-        "php": ">=5.3.3",
+        "php": ">=5.4",
         "civicrm/cv": "~0.2",
         "symfony/console": "~2.8|^3|^4",
         "symfony/process": "~2.8|^3|^4",
         "symfony/templating": "~2.8|^3|^4",
         "symfony/var-dumper": "~2.8|^3|^4",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -4,13 +4,14 @@
...
-        "php": ">=5.3.3",
+        "php": ">=5.4",
         "civicrm/cv": "~0.2",
         "symfony/console": "~2.8|^3|^4",
         "symfony/process": "~2.8|^3|^4",
         "symfony/templating": "~2.8|^3|^4",
         "symfony/var-dumper": "~2.8|^3|^4",
-        "totten/license-data": "dev-master"
+        "totten/license-data": "dev-master",
+        "symfony/var-exporter": "^4.4"
```
totten, line 14: Yeah, the Symfony exporter/dumper are nicer.

Symfony 4.x requires PHP 7.1, so this is effectively raising the minimum PHP version to 7.1.

Which, IMHO, is actually OK on a policy level - ie we're within a couple months of having php71 as an across-the-board minimum. We're not *quite* there yet with 5.21 ESR (PHP 7.0)... but the change is within sight, and `civix` is a dev tool. So let's go ahead and bump it up.

I'm pushing up a few more commits before merging.

PR #189, opened 2020-07-19
Use helper function to get extension directory path

mattwire:
@totten This fixes an issue with the upgrader when run via a core extension - found through eventcart - https://github.com/civicrm/civicrm-core/pull/17339/commits/ad331419797b6071e64d340dd5ca72ad713cc3af

Use the `E::path()` helper to get the path of the extension instead of relying on parent parent parent of current dir..

Also one typo in variable definition.


totten:
Looks good, @mattwire. Did a quick check and confirmed it works the same. 👍

PR #190, opened 2020-08-12
Use single quotes in output as it is best practice to use double only when needed

eileenmcnaughton:
… 

Results in

 const SHORT_NAME = 'deduper';

Not
 const SHORT_NAME = "deduper";


totten:
Yup, good point @eileenmcnaughton. Tested+merging.

PR #191, opened 2020-09-13
Normalize path to xml file in generated DAO files

colemanw:
Overview
----------
Cleans up the "generated from" comment in each DAO file to start from the extension rather than what is often the developer's home directory.

Before
-----------
Comment says `Generated from /home/colemanw/bknix/build/d7/web/sites/default/files/civicrm/ext/org.civicrm.volunteer/xml/schema/CRM/Volunteer/Project.xml`

After
-----------
Comment says `Generated from org.civicrm.volunteer/xml/schema/CRM/Volunteer/Project.xml`


totten:
@colemanw Ah, cool. That looks better to me. :+1: Tested it out -- seems good.

PR #193, opened 2020-09-16
generate:entity - APIv3 "get"ter returns incorrect array

eileenmcnaughton:
See: https://chat.civicrm.org/civicrm/pl/9pf1exq3ypn9ujj153m8tqpkor

@totten @twomice

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -41,6 +41,30 @@ The steps for upgrading the `Upgrader` are as follows:
 
 ## Special Tasks
 
+### Upgrade to v20.09.0+: APIv3 Entity
+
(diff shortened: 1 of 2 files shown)


totten:
Nice find @twomice @eileenmcnaughton.

As I look at this, I feel there are a few additional problems:

1. The entity-name is not escaped. (If you have warnings active, this should complain "Use of undefined constant")
2. We need `UPGRADE.md` instructions
3. When running `tests/make-example.sh`, there is no code-path which actually requests/uses the entity. Consequently, nothing complains if the entity.

I've pushed updates onto this PR for the escaping and UPGRADE.md. To get some coverage, I'll post a separate PR and ping y'all.

twomice:
:+1:

PR #194, opened 2020-09-18
AddEntityCommand - Generate APIv3 unit-test

totten:
This improves test-coverage for APIv3 entities in two ways:

* The command `generate:entity` will create a file `tests/phpunit/api/v3/MyEntityTest.php`.
* The QA script `bash tests/make-example.sh` calls `generate:entity` and subsequently runs `phpunit` on all generated tests. Therefore, it runs

This is motivated by #193, which revealed that `generate:entity` regressed in two ways circa 19.10. I've confirmed that this unit-test would complain about both of the problems.

I've only added the patch for APIv3 (b/c that had the last bug), but an APIv4 patch should look pretty similar and is patch-welcome.

Ping @twomice @eileenmcnaughton @colemanw

### DIFF
--- a/src/CRM/CivixBundle/Command/AddEntityCommand.php
+++ b/src/CRM/CivixBundle/Command/AddEntityCommand.php
@@ -1,6 +1,7 @@
 <?php
 namespace CRM\CivixBundle\Command;
 
+use CRM\CivixBundle\Builder\PHPUnitGenerateInitFiles;
 use CRM\CivixBundle\Services;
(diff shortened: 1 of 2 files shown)


eileenmcnaughton:
Cool - only you can merge...

colemanw:
@totten time for a release?

totten:
@colemanw yup, you read my mind. I just side-tracked because the test-script wasn't passing on Linux (addressed by #195). 

Now published in v20.09.0.

PR #199, opened 2020-10-20
extensions/extensions-directory#18 Update the boilerplate README

mlutfy:
Buncles a few proposed changes to the default README:

* Require PHP 7.2 instead of 7.0, since that is what CiviCRM core recommends
* Cut the installation instructions and refer to the sysadmin guide instead
* Replace "Usage" with "Getting Started" as per https://lab.civicrm.org/extensions/extensions-directory/-/issues/18


totten:
+1 for bumping recommended PHP version and changing header.

Regarding install instructions - I squashed/restored most of the details/examples, but it still adds the link. See #166 which deals with the same thing.

Merging.

PR #200, opened 2020-11-02
Update cv to ~0.3

demeritcowboy:
cv version .2 doesn't work with php 7.4 is the main thing, but also why not.

I'd updated composer.json locally a long time ago for something else that wasn't working - completely forgot about that so I can't say what it was.

e.g. https://civicrm.stackexchange.com/questions/38061/getting-error-when-using-civix-generatemodule-or-civix-generateentity-curly-br

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
     "license": "AGPL-3.0",
     "require": {
         "php": ">=7.1.3",
-        "civicrm/cv": "~0.2",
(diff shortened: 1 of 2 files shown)


totten:
Still works for me on D7 + PHP 7.2 (which is what I happen to have open right now), so... why not merge!

demeritcowboy:
Thanks!

PR #201, opened 2020-11-03
Add apiv4 option to civix:generate api -A 4 Entity Action 

eileenmcnaughton:
I think this needs some more work but it's a start

@colemanw v4api doesn't  lend itself well to adding an action to an existing entity from what I can tell

### DIFF
--- a/src/CRM/CivixBundle/Resources/views/Code/api4.php.php
+++ b/src/CRM/CivixBundle/Resources/views/Code/api4.php.php
@@ -1,0 +1,52 @@
+<?php echo "<?php\n"; ?>
+namespace Civi\Api4;
+
+use Civi\Api4\Generic\AbstractAction;
+use Civi\Api4\Generic\Result;
+use Civi\Api4\Generic\BasicGetFieldsAction;
+
+/**
+*  Class <?php echo $entityNameCamel ?>.
+*
+* Provided by the <?php echo $extensionName ?> extension.
+*
+* @package Civi\Api4
+*/
+class <?php echo $entityNameCamel ?> extends Generic\AbstractEntity {
+
+  /**
+   * <?php echo $entityNameCamel ?> <?php echo $actionNameCamel ?>.
+   *
+   * @return \Civi\Api4\Action\<?php echo $entityNameCamel ?>\<?php echo $actionNameCamel ?>
+   *
+   * @param bool $checkPermissions
+   *
+   * @throws \API_Exception
+   */
+  public static function <?php echo $actionNameLower ?> ($checkPermissions = TRUE): Action\<?php echo $entityNameCamel ?>\<?php echo $actionNameCamel ?> {
+    return (new \Civi\Api4\Action\<?php echo $entityNameCamel ?>\<?php echo $actionNameCamel ?>(__CLASS__, __FUNCTION__))
+      ->setCheckPermissions($checkPermissions);
+  }
+
+ /**
+  * Get permissions.
+  *
+  * It may be that we don't need a permission check on this api at all at there is a check on the entity
+  * retrieved.
+  *
+  * @return array
+  */
+  public static function permissions():array {
+    return ['<?php echo $actionNameLower ?>' => 'administer CiviCRM'];
+  }
+
+ /**
+  * @return \Civi\Api4\Generic\BasicGetFieldsAction
+  */
+  public static function getFields() {
+    return new BasicGetFieldsAction(__CLASS__, __FUNCTION__, function() {
+      return [];
+    });
+  }
+
+}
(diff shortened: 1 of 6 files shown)


colemanw:
@eileenmcnaughton an extension can add an APIv4 action to an existing entity (e.g. an entity declared outside that extension) by providing an action class in the same namespace as that entity. APIv4 will pick it up by magic. E.g.

```
namespace Civi\Api4\Action\Contact;

class Zombify extends \Civi\Api4\Generic\DAOUpdateAction {
```

Then you can call `Civi\Api4\Contact::zombify()->execute()` and it will work. IDEs won't be able to autocomplete that line of code because the function is being called via magic method, but it will show up in the api explorer, etc.

src/CRM/CivixBundle/Resources/views/Code/api4action.php.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+class <?php echo $actionNameCamel ?> extends AbstractAction {
+
+ /**
+  * @inheritDoc
+  *
+  * @param \Civi\Api4\Generic\Result $result
+  *
+  * @throws \API_Exception
+  * @throws \CiviCRM_API3_Exception
```
colemanw, line 22: Why would it throw an `CiviCRM_API3_Exception`?

src/CRM/CivixBundle/Resources/views/Code/api4.php.php
```diff
@@ -0,0 +1,47 @@
...
+use Civi\Api4\Generic\BasicGetFieldsAction;
+
+/**
+*  Class <?php echo $entityNameCamel ?>.
+*
+* Provided by the <?php echo $extensionName ?> extension.
+*
+* @package Civi\Api4
+*/
+class <?php echo $entityNameCamel ?> extends Generic\AbstractEntity {
```
colemanw, line 15: I rarely recommend extending the `AbstractEntity` class directly, as there are more useful base classes.
If the entity in question has a DAO, then it definitely should use `DAOEntity`. But Civix already generates an APIv4 `DAOEntity` whenever generating an entity, so presumably you're targeting a different use-case?

`BasicEntity` is the go-to for any type of CRUD API that isn't backed by a DAO. Is that what you're after?

src/CRM/CivixBundle/Resources/views/Code/api4action.php.php
eileenmcnaughton: copy  & paste - still hitting stuff

totten:
> An extension can add an APIv4 action to an existing entity (e.g. an entity declared outside that extension) by providing an action class in the same namespace as that entity. APIv4 will pick it up by magic.
>
> ... If the entity in question has a DAO, then it definitely should use DAOEntity. But Civix already generates an APIv4 DAOEntity whenever generating an entity, so presumably you're targeting a different use-case? BasicEntity is the go-to for any type of CRUD API that isn't backed by a DAO...

So I guess we have a few use-cases and ideal outcomes:

1. Generate a CRUD entity (which is a different command)
    * `Civi\Api4\MyEntity.php` - **Must** be created
2. Generate a standalone action where there is no pre-existing entity
    * `Civi\Api4\Action\MyEntity\MyAction.php` - **Must** be created
    * `Civi\Api4\MyEntity.php` - **Must** be created.  (At least, I think so. `Civi\Api4\Action\Entity\Get` scans for entities, and it looks for `*.php` entity files in `Civi/Api4/` -- without that, the entity wouldn't seem to exist). This should probably extend `BasicEntity`.
3. Generate a standalone action where the pre-existing entity is in core
    * `Civi\Api4\Action\MyEntity\MyAction.php` - **Must** be created
    * `Civi\Api4\MyEntity.php` - **May** be updated. Ideally, we want this to update in some way (for autocompletion purposes). But if it doesn't, that's OK - it'll still work.
4. Generate a standalone action where the pre-existing entity is in core (or in another ext)
    * `Civi\Api4\Action\MyEntity\MyAction.php` - Needs to be created anew
    * `Civi\Api4\MyEntity.php` ` **Must not** be created. There's already a version of this in core (or in some other extension)

Part of me thinks that maybe we should just skip creating `Civi\Api4\MyEntity.php` so that we don't have to think about `#3`+`#4`. However, this would leave `#2` non-functional. 

So maybe `generate:api -A4` would be something like this pseudocode?

```php
create('./Civi/Api4/Action/MyEntity/MyAction.php');

$entityExists = class_exists('Civi\Api4\MyEntity');
$isMine = file_exists('./Civi/Api4/MyEntity.php');
if (!$entityExists) makeBasicEntity();
elseif ($entityExists && !$isMine) doNothing();
elseif ($entityExists && $isMine) shouldUpdateIfWeFigureOutHowButShrugOtherwise();
```

eileenmcnaughton:
Dang I just created an api for my extension and it gave me v3 & then I realised we never merged apiv4 support to civix (except crud) & ug - it looks like it's still a big lift to get further with it

eileenmcnaughton:
Closing this - I think it has too many gaps

PR #202, opened 2020-11-14
Generate install SQL on the fly

totten:
Overview
----------

This changes the behavior of `generate:entity` / `generate:entity-boilerplate` so that the SQL schema can be adjusted to match runtime conditions. In particular, this aims to allow installation on either `utf8` or `utf8mb4` collations.

Before
-------

1. The command `generate:entity-boilerplate` creates files `sql/auto_install.sql` and `sql/auto_uninstall.sql`.
2. During installation/uninstallation, the `CRM_*_Upgrader` executes these (via `onInstall()` and `onUninstall()`).

After
-------

1. The command `generate:entity-boilerplate` creates files `sql/create_example.sql` and `sql/drop_example.sql`. These allow you to preview/inspect the schema, but they are inert.
2. During installation/uninstallation, the `CRM_*_Upgrader` generates fresh SQL. Specifically, `onInstall()` and `onUninstall()` use the "schema builder" to generate code.

Comments
-----------

(A) It would be problematic if you have *both* `auto_install.sql` *and* "schema builder" calls -- ie it would run `CREATE TABLE` twice.

(B) The command `generate:entity-boilerplate` still serves the same general purpose in the dev-workflow. However, it does some sanity-checks to help the transition:

Screen Shot 2020-11-14 at 2 50 43 PM

(C) I haven't actually tested the "schema builder" with different collations (`utf8`/`utf8mb4`).

(D) Right now, it embeds a copy of the "schema builder" in the `CRM_*_Upgrader_Base`. Short-term, this allows easier compatibility/interoperability between different versions of extensions+core. Long-term, it might be better if this migrated to core. If we want to aim for that, then perhaps path should be:

* Add a similar "schema builder" to `civicrm-core`@`master`
* Update this PR -- the `createSchemaBuilder()` should first see if we have a class from core. Use that if available. Otherwise, use the embedded fallback version.
* Merge PR and release next version of `civix`

(E) This patch-set includes some refactoring that makes it easier to work on the templates. Specifically, the IDE can be futzy about formatting `*.php.php` files.  the `CRM_*_Upgrader_Base` is a good example where we don't need full dynamism of `*.php.php`; instead, we can just take a model `*.php` file and do search/replace. The `*.php` file is more amenable to editing in an IDE.


mattwire:
@totten I think @artfulrobot and @mikeymjco will have some thoughts on this.

First, I think it's a good idea to autogenerate schema, that allows adaptation and also avoids the issue of xml being out of sync with SQL.

My thoughts are:
* This seems like a lot of extra code in order to support the automatic generation of schema.
* I don't think we should be adding code to support < CiviCRM 5.23.
* Actually writing SQL files on install will be problematic if the extensions dir is readonly.
* Given that full support for utf8mb4 has only really been around since around 5.31 couldn't we just add something to core and not bloat the extensions. I can see the ability to autogenerate sql as a core function being very useful for other things too.
* This is mostly a configuration issue that can easily be avoided by not specifying the character set in the sql - see https://lab.civicrm.org/extensions/stripe/-/blob/master/sql/customers_install.sql. I realise that some hosting may not have the "right" default setup but they'll probably run into other problems too in that case.

eileenmcnaughton:
So the main concern appears to be bloat - an extra 500~ lines.

I note the plan of

> (D) Right now, it embeds a copy of the "schema builder" in the CRM_*_Upgrader_Base. Short-term, this allows easier compatibility/interoperability between different versions of extensions+core. Long-term, it might be better if this migrated to core. If we want to aim for that, then perhaps path should be:
> 
> Add a similar "schema builder" to civicrm-core@master
> Update this PR -- the createSchemaBuilder() should first see if we have a class from core. Use that if available. Otherwise, use the embedded fallback version.
> Merge PR and release next version of civix

So it seems this would be added to civix / new extensions now and over the next x months it would become obsolete again.

I wonder if we could put more control or at least more information into the extension writer's hands. For example the way I manage upgrades I would simply tag a new release for my extensions that only supports 5.33 + and no fallback would be required - so anyone on a version of one of my extensions that supports this install script would also be on a version of core that has the class - making it unnecessary for all of my extensions - so if there were a switch ---skip-fallback-install I would use that. I imagine that would increasingly be the case as 5.33 moves into the rear-view mirror but it might be handy for sometime for those extensions that choose to support dozens of versions to get this fallback.

I suspect the bloat within civix iteself is not really a concern but we could perhaps add some comments to indicate that the fallback would be removed at some point in the future - perhaps 18 months away

eileenmcnaughton:
@mattwire I don't think the sql is written to disk so the file permissions shouldn't matter?

seamuslee001:
In regards to the not specifying the Character Set @mattwire @totten I think what that would mean is it falls back to the database default right? This might be potentially problematic on a shared host or in the case where say you have CiviCRM tables and CMS tables installed on the one DB and CiviCRM have been switched but not the CMS's and the DB default is utf8 rather than utf8mb4?, I still wonder if we can get away with using the auto_install.sql approach but instead inject something that does a SELECT for the character set etc and use MySQL variables in the resulting create table?.

eileenmcnaughton:
@seamuslee001 we did test using mysql variables (I thought you were on that call?) and found we couldn't use mysql variables. Once we were into smarty variables we wer e already into generated code & parsing smarty - at which point this started to look a whole lot cleaner.

@artfulrobot I think an extension that wanted something more nuanced still has several options for overriding the generated function - ie the install hook can just be edited to call a mysql file. 

The goal is that most of the time we would be creating tables off the xml that can install with the right charset even when the hosting provider locks down the ability to change the db charset. When the extensions want to over-ride this behaviour they can - but our goal is to make it easy for extension writers doing normalish stuff without preventing them  from doing weird & wonderful stuff if necessary.

seamuslee001:
@eileenmcnaughton I thought we had only discussed MySQL variables not done any actual testing

eileenmcnaughton:
@seamuslee001 no @totten & I tested it while you thought we were  listening to you :-)

src/CRM/CivixBundle/Resources/Example/SchemaBuilderTrait.php
```diff
@@ -0,0 +1,256 @@
...
+        $config = new \stdClass();
+        $config->sqlCodePath = dirname($tmpFile);
+        $config->tables = $this->tables;
+
+        $schema = new \CRM_Core_CodeGen_Schema($config);
+
+        // We're poking into an internal class+function (`$schema->$generator()`) that changed in v5.23.
+        // Beginning in 5.23: $schema->$function() returns an array with file content.
+        // Before 5.23: $schema->$function($fileName) creates $fileName and returns void.
+        if (version_compare(\CRM_Utils_System::version(), '5.23.alpha1', '>=')) {
```
mattwire, line 158: Do we really need a compatibility layer for 5.23? Given that 5.27 is ESR and 5.28 is security release

totten: @mattwire Yeah, it seems borderline to me.

* Agree - the fact that 5.23 predates the current 5.27 ESR is an indicator that it can be removed. 
* OTOH, it 5.23 is less than a year old (March 2020).
* Removing the guard would mean that an extension using `generate:entity-boilerplate` would need to bump up its minimum. (The current default is `<compatibility><ver>5.0</ver></compatibility>`.)
* A lot of folks who write extensions are a few versions back.
* Basically, from my POV, the existing code already showed how to support <5.23, and it seemed easier to keep the compatibility than manage the removal.
* Then again, it should be removed at some point. If not now - then when?
* Updating `civix.git` doesn't automatically break existing extensions. The downstream dev needs to get the newer version and run it. If you need backward compat and if you're conscientious, then there are ways to proceed (e.g. stick to an old version of civix; e.g. play close attention to the diff's).
* It's not easy to download an older version of `civix`. There's only one published URL (`https://download.civicrm.org/civix/civix.phar`) which is evergreen. The versioned URLs aren't publicly visible.
* If we published/hyperlinked the older builds, then maybe we could be more aggressive about phase-outs within `civix`.

eileenmcnaughton: I'm not sure I would have written the support for pre 5.23 - but given that it is written I think we should keep for now

eileenmcnaughton:
@totten @seamuslee001 @mikeyMJO @mattwire  @artfulrobot so to come back to where this is at - I feel convergence on 

1) just - at a high level this makes sense but
2) we should plan to move the helper functions to core and
3) we should ensure it's documented how extensions can create tables in 'whatever way they want'

So the only blocker at this point seems to be around documentation - does that sound right to people?

src/CRM/CivixBundle/Resources/Example/SchemaBuilderTrait.php
```diff
@@ -0,0 +1,256 @@
...
+          return array_shift($schema->$generator());
+        }
```
demeritcowboy, line 160: I was playing with this trait a bit to see if I could use it standalone. Partial success but this line didn't work and it seems like it should always fail?? `array_shift()` uses pass by reference, so it needs an actual variable.

The main other thing that came up is there's no way to set $database and it doesn't seem to set it internally either, so it errors in a couple places.

colemanw:
@totten this needs rebase

totten:
This branch included a bunch of refactoring. There might be some use in re-reading those parts, and some of those ideas have taken other forms now - so it needs a rethink, and I wouldn't want to specifically build on this branch. No point keeping the PR open.

Issue #206, opened 2021-02-24
Deprecated: Array and string offset access syntax with curly braces is deprecated


adixon:
Extensions using civix are generating the above deprecation notice (as of civicrm 5.34.0, at least).

The offending code is just some old backward compatibility stuff to support CiviCRM versions less than 4.3.

https://github.com/totten/civix/blob/f695795a51ce9d9e36f6b6029a3779856642a8ec/src/CRM/CivixBundle/Resources/views/Code/module.civix.php.php#L234

PR forthcoming.

eileenmcnaughton:
@adixon I don't have rights to close this but looks like a fix was merged

adixon:
Yup!

PR #207, opened 2021-02-24
remove old backward compatibility code that generates deprecation warnings

adixon:
As per https://github.com/totten/civix/issues/206


totten:
I suppose we don't need to worry about compatibility with <4.3 :smile: 

Good simplification. Did a sanity check on 5.35 - test output was good. Merging.

PR #209, opened 2021-03-17
PHPUnit 8 - test-api.php.php

eileenmcnaughton:


seamuslee001:
LGTM @totten

src/CRM/CivixBundle/Resources/views/Code/test-api.php.php
```diff
@@ -21,7 +21,7 @@ class <?php echo $testClassName ?> extends \PHPUnit\Framework\TestCase implement
    *
    * See: https://docs.civicrm.org/dev/en/latest/testing/phpunit/#civitest
    */
-  public function setUpHeadless() {
+  public function setUpHeadless(): void {
```
totten, line 24: Doesn't make sense on a few fronts:

* The function here actually does return a value, which raises a PHP error.
* This disagrees with `HeadlessInterface`'s signature for `setUpHeadless()`
* Because the templates have it return a value... therefore one should expect extant tests to also return a value.

totten, line 24: Doesn't make sense on a few fronts:

* The function here actually does return a value, which raises a PHP error.
* This disagrees with `HeadlessInterface`'s signature for `setUpHeadless()`
* Because the templates have it return a value... therefore one should expect extant tests to also return a value.

eileenmcnaughton:
Closing this as easier to re-do than fix I think & the other one is good-to-merge

PR #214, opened 2021-03-17
PHPUnit 8 - test-headless.php.php

eileenmcnaughton:


src/CRM/CivixBundle/Resources/views/Code/test-headless.php.php
```diff
@@ -28,34 +29,43 @@
...
+   * 
+   * Civi\Test has many helpers, like install(), uninstall(), sql(), and sqlFile().
+   * 
+   * @see See: https://github.com/civicrm/org.civicrm.testapalooza/blob/master/civi-test.md
+   *
+   * @return \Civi\Test\CiviEnvBuilder
+   *
+   * @throws \CRM_Extension_Exception_ParseException
+   */
+  public function setUpHeadless():void {
```
seamuslee001, line 43: not sure if this is needed as its a internal Civi only thing but probably good to do

seamuslee001:
LGTM @totten

src/CRM/CivixBundle/Resources/views/Code/test-headless.php.php
totten: Doesn't make sense on a few fronts:

* The function here actually does return a value, which raises a PHP error.
* This disagrees with `HeadlessInterface`'s signature for `setUpHeadless()`
* Because the templates have it return a value... therefore one should expect extant tests to also return a value.

eileenmcnaughton: updated the return to   ```CiviEnvBuilder```

eileenmcnaughton:
@totten  - looks like this was fixed up a couple of months back

seamuslee001:
@totten  I manually applied patch locally and for an extension I work with I manually generated a FootTest Class and then ran phpunit and the test suite ran successfully

totten:
Brilliant. Thanks @seamuslee001

PR #215, opened 2021-03-19
Remove safe-mode as it is php5.4

eileenmcnaughton:
Other minor fixups to preferred php (as per inspections) which I opened this to demonstrate https://github.com/civicrm/civicrm-core/pull/19834

### DIFF
--- a/src/CRM/CivixBundle/Resources/views/Code/phpunit-boot-cv.php.php
+++ b/src/CRM/CivixBundle/Resources/views/Code/phpunit-boot-cv.php.php
@@ -3,7 +3,7 @@ echo "<?php\n";
 ?>
 
 ini_set('memory_limit', '2G');
-ini_set('safe_mode', 0);
+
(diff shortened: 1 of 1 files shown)


seamuslee001:
LGTM @totten

src/CRM/CivixBundle/Resources/views/Code/phpunit-boot-cv.php.php
```diff
@@ -27,11 +27,11 @@
  * @throws \RuntimeException
  *   If the command terminates abnormally.
  */
-function cv($cmd, $decode = 'json') {
+function cv(string $cmd, $decode = 'json'): string {
```
totten: The return type here is not correct -- if the `$decode` mode is `raw` or `phpcode`, then it will return `string`. But if it's `json` mode, then it can be any JSON-style value. Pushed a correction.

PR #225, opened 2021-07-15
Add typing to upgrader functions

eileenmcnaughton:
This will help when diagnosing a fairly common gotcha - ie forgetting to return TRUE.

Currently there are very few clues as to when that is the case - so if these
type hints become part of the copy & paste it will help with that

@ejegg


totten:
Looks good -- (1) more accurate type hint, (2) only updates comments/suggestions, (3) the relevant functions are invoked dynamically, which means there's no concern about (mis)matching formal signatures.

totten:
> t would also be good to update the docblock because `@return TRUE on success` is not a valid annotation

Makes sense.

For updating this template, we should just move the explanation to the next line

```php
 * @return bool
 *   TRUE on success.
```

I know that many folks would infer that `@return bool` in this context is a success/failure indication, but it's not a universal convention --  better to be explicit. Part of the role of the boilerplate is to educate.

PR #227, opened 2021-09-25
Allow underscores in module names

colemanw:
Previously the command `civix generate:module "search_kit"` would throw `'Malformed package name'`

Note: I moved the `-` character to the beginning of each `[]` block for readability; dashes do not need to be escaped if they are the first character.


totten:
Change looks about right. I'll try to r-run when I get back to the computer later today.

colemanw:
Changes look good.

totten:
Did some light `r-run` and `test/make-example.sh`. Pushed a couple tweaks onto this branch. The stuff in worked as expected.

I needed a couple other patches for tangential issues, and I'll just push those up directly after merging this.

There is a related issue - if one actually uses underscores, then all the `generate:angular*` commands pick wonky names for the Angular module. I'll open a PR for that soon after. For now, we can merge this.

PR #228, opened 2021-09-26
Angular - Default module-name should be well-defined in an `_`-world

totten:
This is a followup to #227, which allows use of `_` in module-names. If you have underscore names and then generate Angular code, the module names are peculiar. This updates the naming convention. Below, I've reproduced the explanation from `UPGRADE.md`:

ping @colemanw 

--

When generating Angular code, the default Angular module name has traditionally matched the extension's short-name, eg

| __Long Name (Key)__ | __Short Name (file)__ | __Default Angular Module__ |
| -- | -- | -- |
| `org.example.foobar` | `foobar` | `foobar` |
| `org.example.foo-bar` | `foo_bar` | `foo_bar` (!!) |

This has two problems:

* If the extension has dashes or underscores, then the Angular module winds up with underscores. Underscores
  do not map sensibly to HTML tags or attributes.
* There is no standard prefix. By convention, most CiviCRM Angular modules use the prefix `crm` (`crm-`).

In v21.09.0+, the `info.xml` may explicitly specify the Angular module, as in:

```xml
<extension key="org.example.foo-bar" type="module">
  <file>foo_bar</file>
  <civix>
    <namespace>CRM/FooBar</namespace>
    <angularModule>crmFooBar</namespace>
  </civix>
</extension>
```

For new extensions, the `info.xml` will be set the name as follows:

| __Long Name (Key)__ | __Short Name (file)__ | __Default Angular Module__ |
| -- | -- | -- |
| `org.example.foobar` | `foobar` | `crmFoobar` |
| `org.example.foo-bar` | `foo_bar` | `crmFooBar` |

Without a directive in `info.xml`, the default name will continue as before. However, you may wish to update some existing extensions to explicitly set the preferred name in `info.xml`.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -41,42 +41,28 @@ The steps for upgrading the `Upgrader` are as follows:
 
 ## Special Tasks
 
-### Upgrade to v21.09.0+: Angular Module
+### Upgrade to v21.09.0+: Angular Module (optional)
 
-When generating Angular code, the default Angular module name has traditionally matched the extension's short-name, eg
+Angular code in Civi extensions usually has one of these layouts:
 
-| __Long Name (Key)__ | __Short Name (file)__ | __Default Angular Module__ |
-| -- | -- | -- |
-| `org.example.foobar` | `foobar` | `foobar` |
-| `org.example.foo-bar` | `foo_bar` | `foo_bar` (!!) |
+* (A) (default, most common) There is one Angular module, and its name exactly matches the Civi extension.
+* (B) There is one Angular module, and its name does *not* match the Civi extension.
+* (C) There are multiple Angular modules. It is impossible for them to all match.
 
-This has two problems:
-
-* If the extension has dashes or underscores, then the Angular module winds up with underscores. Underscores
-  do not map sensibly to HTML tags or attributes.
-* There is no standard prefix. By convention, most CiviCRM Angular modules use the prefix `crm` (`crm-`).
-
-In v21.09.0+, the `info.xml` may explicitly specify the Angular module, as in:
+Extensions with mismatched names (group B) may now provide a hint via `info.xml`. For example, if the extension is `foobar` and the Angular module is `crmFoobar`, then set:
 
 ```xml
-<extension key="org.example.foo-bar" type="module">
-  <file>foo_bar</file>
+<extension key="com.example.foobar" type="module">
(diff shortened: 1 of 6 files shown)


colemanw:
I'm not completely sold on declaring `<angularModule>` on a per-extension basis, because most of the extensions I write contain more than one Angular module.

Maybe my use-cases are the outlier though, and most people trying to spin up an Angular module in an extension are just wanting to create a single customization rather than a multi-featured app.

I understand wanting to create a more sensible default module name, but I don't understand why that default would need to be stored in the XML rather than just be calculated at runtime. It seems counter-intuitive to me that when using the `civix generate:angular-module` command I would need to edit the xml to specify a module name; why not just pass it in as an arg on the command line if I want to customize it?

totten:
> why not just pass it in as an arg on the command line if I want to customize it?

That still works with this patch (ie `--am=...`).  The `info.xml` changes the default/implicit module.  So if one prefers to do multiple modules, that's just as easy/hard as it is today. I've rewritten the `UPGRADE` notes to clarify that (1) it's a very small group of extensions that would want to edit `info.xml` and (2) even then, it's optional - it's a usability tweak.

>  I don't understand why that default would need to be stored in the XML rather than just be calculated at runtime

Initially, I did that - update the runtime calculation to camelize names. But then I realized that this would break codegen on existing extensions that already have underscores in their name. (And, it turns out, there are a few of them in `universe`. The underscore is technically legal for Angular modules -- the problem is that it goes against the broader convention of camelcase-ing Angular artifacts.)

In any event, the `info.xml` bit was initially an offramp for extensions with underscores, so I started writing (`UPGRADE.md`) a longish explanation of this offramp. But it sucks to be required to update `info.xml` on existing extensions. Do really we need to?

It struck me that we'd have simpler transition if the new/opinionated formula was limited to `InitCommand`. With this arrangement:

* We can keep the runtime formula -- so all existing extensions continue to work as before.
* We can change the init-time formula more freely -- eg we can add the `crm` prefix to new extensions without breaking existing ones. Six months later, we could change it again.
* We can improve compat with existing oddballs like `mosaico` and `extensionui`. (They each have one Angular module - but the module-name is unconventional.)

colemanw:
Ok, that makes sense. I don't have write access to this repo, but :+1: from me to merge this.

PR #230, opened 2021-10-27
refs #composer updated Symfony components (^5)

kewljuice:
Updated Symfony components so requirements for composer are met.


totten:
Merging. I haven't really tested, but I'm doing a bunch of generalized testing for some other revamps, and it's probably ideal to get this kind of change in at the start of nontrivial dev-test loop.

Issue #175, opened 2020-02-28
Add CiviX class in Core?


artfulrobot:
Every extension created ends up with a `myext.civix.php` file which is all duplicate boilerplate code.

I was wondering if it would make more sense to define a Civi\Civix (or some other name) class that's used instead of the duplicated global functions?

totten:
I like the idea of cleaning this up. I don't think I have a strong enough sense of what you're describing to agree or disagree on defining it as a class... but let me think out loud a moment...

The basic reason for the existence of `*.civix.php` file (and all the `_civix_` functions) that `civix` expresses a different opinion on how to use various hooks/APIs. Compare these opinions:

1. Downstream authors should register routes by creating an XML file and creating a PHP function which returns the name of the XML file. The PHP function function should follow a naming-convention.
2. Downstream authors should register routes by running a command which creates an XML file. The XML file should follow a naming-convention.

`Opinion #1` is supported by `civicrm-core`. `Opinion #2` is supported by `civix`, and `civix` provides a bridge/shim (via `*.civix.php` boilerplate and `_civix_` functions).

At the time when the bridge/shim structure was designed, a few factors weighed in - factors which have changed.

* __Limits in hook system__: The *only* way to tap into CiviCRM hooks was to create boilerplate hooks - so everything had to flow from there.
* __Lifecycle__: A downstream administrator installs some specific version of `civicrm-core` and runs that version for a long time (like 6-18 months), which means it is locked into `Opinion #1`. During that time, there may be updates in the developer-tooling (providing the better `Opinion #2`) - but the only way to go-live with `Opinion #2` is to have a shim/bridge/polyfill for use by the extension. (If you tried to put `Opinion #2` directly in `civicrm-core`, then you'd be waiting at least 6 months for a stable release.)

But now:

* We have the option to register hooks as Symfony listeners, which are a bit more open/programmable than the function-naming-convention.
* The cadence of `civicrm-core` is faster than it used be -- ie if you put `Opinion #2` directly in `civicrm-core`, it will come out in 1-2 months instead of 6-8 months.

Tangentially, the current bridge/shim design has always had a few annoyances:

* If a bridge/shim needs an update, then every downstream dev has to regen `mymodule.civix.php`.
* If a new bridge/shim needs to be added, then each downstream dev has to regen `mymodule.civix.php` **and** manually edit the main `mymodule.php` to register them. (Witness: `UPGRADE.md`.)
[...]

PR #233, opened 2021-12-08
(#175) Generate code with mixins instead of hook boilerplate

totten:
* Basline support for loading mixins
    * [x] Download backported mixin files
    * [x] Add helper to copy backported mixins into extensions
* Assimilate specific mixins (update code-generators; transition boilerplate)
    * [x] `mixin/theme-php`
    * [x] `mixin/menu-xml`
    * [x] `mixin/case-xml`
    * [x] `mixin/setting-php`
    * [x] `mixin/mgd-php`
    * [x] `mixin/ang-php`
* Assimilate other modernizations
    * [ ] `<classloader>`
    * [ ] `<upgrader>`
* Generic support
    * [x] `civix generate:module` - Add options to initialize some default mixins on new exts
    * [x] `civix upgrade` - Add command to perform/recommend various updates to `info.xml`, `mymod.php`, etal
    * [x] `civix mixin:*` - Add commands to enable/disable/list mixins
* Documentation
    * [ ] Introduce mixins
    * [ ] Revise generator-specific docs/examples
    * [ ] Upgrading


totten:
@colemanw I've added the command `civix upgrade`. Running this should do the needful:

* Enable the mixins if needed (eg enable `mgd-php` if you have any `*.mgd.php` files)
* Remove calls to the dead functions (eg `_*_civix_civicrm_managed()`) from `mymodule.php`
* Regenerate the `mymodule.civix.php` (with less boilerplate)

In particular, note the files:

* The "2.0" upgrade script, which enables mixins and removes dead function-calls: https://github.com/totten/civix/blob/master-mixin/upgrades/2.0.up.php
* The utility class, which defines helpers for applying changes: https://github.com/totten/civix/blob/master-mixin/src/CRM/CivixBundle/Upgrader.php

There are a few things which would nice -- but which I didn't do. These are just a bit more work, and they're not *critical*  enabling mixins. They can be done as separate updates (eg a `2.1` update):

* Remove empty stubs (eg after removing `_foobar_civix_civicrm_managed()`, the higher-level  `foobar_civicrm_managed()` could be empty). However, it should work fine, and it should be fairly obvious when skimming the diff.
    * A good way to go at this might be to add a helper `$upgrader->removeEmptyFunctions()` or `$upgrader->removeIfEmpty($functionName)`.
* In the `info.xml`, setting up `<classloader>` and `<upgrader>`. I mentioned this at the top, but that improvement is independent of the mixin changes.

colemanw:
@totten side-note, I think we should stop generating code for `hook_civicrm_casetypes()` from Civix.
See https://github.com/civicrm/civicrm-core/pull/23320

totten:
I did a bunch of testing yesterday -- eg using several versions of Civi (5.39, 5.44, 5.45, master) to run `tests/make-example.sh` (which exercises various code-generators and runs the auto-generated items) and to manually access some of the generated resources. All passed.

Additionally, I've locally tried the `civix upgrade` on a few extensions:

* `mosaico`
* a new/throw-away extension initialized with a much older version (civix 16.03)
* a new/throw-away extension initialized with a recent version (civix 22.02)
* `search_kit

> (@colemanw) side-note, I think we should stop generating code for hook_civicrm_casetypes() from Civix.

I'm OK with that. I don't feel a rush to try to squeeze that into this PR (ie there's a lot of other stuff wrapped up in here, and I don't think this has a decisive effect on transitioning `hook_caseTypes`=>`hook_managed`). I think there might be a good story to tell about how this allows more targeted deprecations (ie if all extensions implement `hook_caseTypes`, then there's more noise to sift-through when removing it from all extensions).

PR #235, opened 2022-01-26
Update config function

eileenmcnaughton:
I'm getting notices about this line
```
  $template = & CRM_Core_Smarty::singleton();
```

PHP Notice:  Only variables should be assigned by reference in /srv/civi-sites/wmff/civicrm/ext/greenwich/greenwich.civix.php on line 94

__DIR__ is just a teensy bit shorter / more readable


colemanw:
@totten I don't have merge rights here - can you please merge this

PR #241, opened 2022-05-19
Add extension dirs to class loader in bootstrap.php template

dontub:
To run unit tests that do not require a CiviCRM environment the extension's classes need to be loadable by the class loader.

It might be considered to do this for every extension in case the extension under test is in the `ext` directory so that it is possible to access classes of other extensions.

### DIFF
--- a/src/CRM/CivixBundle/Resources/views/Code/phpunit-boot-cv.php.php
+++ b/src/CRM/CivixBundle/Resources/views/Code/phpunit-boot-cv.php.php
@@ -9,10 +9,12 @@ eval(cv('php:boot --level=classloader', 'phpcode'));
 // phpcs:enable
 // Allow autoloading of PHPUnit helper classes in this extension.
 $loader = new \Composer\Autoload\ClassLoader();
-$loader->add('CRM_', __DIR__);
-$loader->add('Civi\\', __DIR__);
-$loader->add('api_', __DIR__);
-$loader->add('api\\', __DIR__);
+$loader = new \Composer\Autoload\ClassLoader();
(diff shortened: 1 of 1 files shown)


totten:
Thanks @dontub. That's a really good point about *unit* tests (as opposed to the [headless and e2e tests](https://docs.civicrm.org/dev/en/latest/testing/) which are prevalent in  `civicrm-core.git:tests/`).

Overall, I like this. Some back/forth thoughts:

* __Good__: There may be some upside in encouraging *unit* tests - headless+e2e are not necessarily the gentlest way to get started. Maybe (at some point) this leads to an easier starting-point (where a developer's first test is a *unit* test that runs in a thinner/simpler environment).
* __Bad__: There may be some edge-cases where this reduces the fidelity of headless tests. (*Ex: If someone deletes the `<classloader>` tag, then a headless test should show true-failure due to classloading bug; but the extra classloader here could cover up the bug -- leading to a false-success.*) OTOH, almost everybody uses the default classloading rules; almost nobody customizes them; and this matches the default well. In practice, it should still be representative.
* __Good__: There is a benefit on a related issue (*something that has bitten a few people, including me*) -- namely, can phpunit tests use base-classes or traits from `./CRM` or `./Civi`? This revision should make the system more permissive/forgiving about when you can pull in helpers.
* __Good__: I did a run with `tests/make-example.sh` and confirmed that the example tests worked just as well before+after the 
patch.

So... Let's give it a try!

(As a side note -- I sometimes find myself wishing for a stricter separation between the suites; eg maybe separate `phpunit.xml` and separate `bootstrap.php` for each suite. But that's a bigger change.)

Issue #221, opened 2021-04-23
`civix generate:upgrader` should use core upgrader (5.38+)


totten:
This a follow-up to https://lab.civicrm.org/dev/core/-/issues/2547 and https://github.com/civicrm/civicrm-core/pull/20090. Beginning with civicrm-core in v5.38+, the `CRM_*_Upgrader_Base` will be available as `CRM_Extension_Upgrader_Base`. This means that:

* We won't need so much boilerplate.
* Upgrade steps are sorted on extension dependencies (more primitive ext's are upgraded before more dependent ext's)
* It will be easier to do patches/fixes on the base upgrader (e.g. if we need to distinguish the upgrade steps defined by each ext).

However, extensions will need a transition. For a manual transition, the steps are:

1. Edit `CRM/Myext/Upgrader.php`. Change the base-class from `CRM_Myext_Upgrader_Base` to `CRM_Extension_Upgrader_Base`.
2. Edit `info.xml`.
    * Ensure that `<classloader>` has a `<psr0>` statement
    * Ensure that `<upgrader>` points to `CRM_Myext_Upgrader`
3. Delete `CRM/Myext/Upgrader/Base.php`
4. Remove old stubs from `myext.php` (ie `myext_civicrm_{install,disable,enable,uninstall,upgrade}`)
5. Optionally, remove old stubs from `myext.civix.php` (ie `myext_civicrm_{install,disable,enable,uninstall,upgrade}`)

The command `civix generate:upgrader` should be revised to:

* Generate new/future upgraders based on this convention
* Update existing upgraders based on this convention

Ideally, if someone wants to stick with the old upgrader for a while (e.g. because they need to run on <v5.38), then that should be possible. This probably affects item `#5` above -- we should continue providing stubs in `myext.civix.php` during a transitional period.

mattwire:
@totten I wonder if there is any way of auto-checking any of this, similar to the core code style checker?

There are some "checks" that I automatically do every time I touch an extension such as check for postInstall stub (if missing upgrader runs on install), regenerate civix etc. and the list above adds a whole load more things to check.

totten:
@mattwire Yeah, some linting would be cool.

I guess in an ideal world we would update https://github.com/civicrm/coder with some hints (e.g. "if `CRM_(.*)_Upgrader` extends eponymous `CRM_(.*)_Upgrader_Base`, then show a warning about the preferred subclass)? I suppose there's a learning-curve to writing the sniffs for phpcs...

[...]

PR #245, opened 2022-06-17
upgrades/16.10.0.up.php - Don't add hook_postInstall if `<upgrader>` registered via XML

totten:
Extensions going from (say) v15.* to v22.05 should add `hook_postInstall` (as in `16.10.0.up.php`)... unless they have been manually edited to use `<upgrader>`.

CC @colemanw 
See: https://github.com/totten/civix/issues/221#issuecomment-1157599835


colemanw:
Looks good to me @totten. I don't have permission to merge.

totten:
Published as v22.06.0.

Issue #237, opened 2022-03-25
civix generate:entity-boilerplate Not generating SQL in correct order for dependancies


rbaugh:
The [oderTables](https://github.com/totten/civix/blob/master/src/CRM/CivixBundle/Command/AddEntityBoilerplateCommand.php#L152) function isn't making sure table referenced for FK are ordered first so that when the SQL is generated, the tables are created in a fashion that the constraint can be set as needed for the FK.

The issue comes in that the order the tables come into the function, from the [glob](https://github.com/totten/civix/blob/master/src/CRM/CivixBundle/Command/AddEntityBoilerplateCommand.php#L63) call, essentially dictate how the SQL will be generated. If the order of the files, based on the glob order, is off then the orderTables function just iterates over that oder.

Steps to reproduce.
1. Use civix generate:entity to create two tables, Type and Report
2. In the Report XML file, create a field that is a FK to the Type.ID table.
3. Run civix generate:entity-boilerplate

When you check the sql in the auto-install.sql file, you will see that the Create Report table is listed first and therefore will throw an error when trying to run the SQL file.

If you rename the Report.xml file to _Report.xml, and re-run civix generate:entity-boilerplate, you will see the SQL code is in the correct order now that the files are in the correct order for the loop in [orderTables](https://github.com/totten/civix/blob/master/src/CRM/CivixBundle/Command/AddEntityBoilerplateCommand.php#L152)

Talking with @colemanw, he mentioned that [Afform](https://github.com/civicrm/civicrm-core/blob/5c89828ccfa4f48906ad041b68cebccbb736849d/ext/afform/core/Civi/Afform/Utils.php#L22-L50) and [Extension manager](https://github.com/civicrm/civicrm-core/blob/9b500da6a0d67208c07e740bab88d27abd469d4c/CRM/Extension/Manager.php#L752) are using TopSort library which could be an optional upgrade for the sorting.

colemanw:
Note: there is a function in CiviX called "OrderTables" which I haven't looked at closely but one would think its job was to, you know, order the tables:
https://github.com/totten/civix/blob/063e4139b9ed5466fb747ed1421cbfeb4f6e121a/src/CRM/CivixBundle/Command/AddEntityBoilerplateCommand.php#L152-L174

aydun:
@rbaugh - Could you test this PR please?  https://github.com/totten/civix/pull/247

PR #247, opened 2022-07-08
Fix https://github.com/totten/civix/issues/237

aydun:


colemanw:
Cool. What do you think @totten?

totten:
Nicely done @aydun!

I added a basic test case, and the patch appears to fix the ordering. (Expected ordering: `Flour` before `Bread`; `Bread` before `Sandwich`; `Sandwich` before`Meal`)

@michaelmcandrew originally wrote this algorithm, so he might have more thoughts on it. Let's wait a couple days to see if he's got feedback.

aydun:
Nice!  Looks like there was more work required for the test than the fix!

totten:
Cool, merging. Thanks all!

totten:
Released in v22.07.0.

Issue #249, opened 2022-07-19
generate:entity-boilerplate errors output in auto_install.sql


rbaugh:
So this is somewhat in reply to ticket #237. I didn't get a chance to test the patch before it was merged but have tested and started to get some odd errors. I am getting the following error code dumped into my auto_install.sql file.

```
Warning: Trying to access array offset on value of type null in /private/var/folders/bn/zjh6cq791rb9n1gb0p62m1v00000gn/T/templates_c_QEmK7y.d/%%29^298^298C3DBD%%schema.tpl.php on line 17

Call Stack:
16788509087.6185     356056   1. {main}() /usr/local/bin/civix:0
16788509087.6185     594656   2. require('phar:///usr/local/bin/civix/bin/civix') /usr/local/bin/civix:14
16788509087.6185     673272   3. CRM\CivixBundle\Application::main() phar:///usr/local/bin/civix/bin/civix:28
16788509087.6185    1056000   4. CRM\CivixBundle\Application->run() phar:///usr/local/bin/civix/src/CRM/CivixBundle/Application.php:38
16788509087.6185    1149840   5. CRM\CivixBundle\Application->doRun() phar:///usr/local/bin/civix/vendor/symfony/console/Application.php:149
16788509087.6185    1153224   6. CRM\CivixBundle\Application->doRunCommand() phar:///usr/local/bin/civix/vendor/symfony/console/Application.php:273
16788509087.6185    1153224   7. CRM\CivixBundle\Command\AddEntityBoilerplateCommand->run() phar:///usr/local/bin/civix/vendor/symfony/console/Application.php:1009
16788509087.6185    1155056   8. CRM\CivixBundle\Command\AddEntityBoilerplateCommand->execute() phar:///usr/local/bin/civix/vendor/symfony/console/Command/Command.php:255
16788509087.6185   43879200   9. CRM\CivixBundle\Command\AddEntityBoilerplateCommand->CRM\CivixBundle\Command\{closure:phar:///usr/local/bin/civix/src/CRM/CivixBundle/Command/AddEntityBoilerplateCommand.php:118-137}() phar:///usr/local/bin/civix/src/CRM/CivixBundle/Command/AddEntityBoilerplateCommand.php:138
16788509087.6185   43879392  10. CRM_Core_CodeGen_Schema->generateCreateSql() phar:///usr/local/bin/civix/src/CRM/CivixBundle/Command/AddEntityBoilerplateCommand.php:125
[...]

PR #251, opened 2022-07-26
AddEntityBoilerplateCommand - Fill in the `$config->database` properties

totten:
Overview
----------

This aims to address the errors outputted by `generate:entity-boilerplate` command (as reported by @rbaugh in #249). Errors are reproducible on php80+WordPress (with more steps described in #249). The errors indicate that some configuration-options are unavailable. The patch provides the missing options.

cc @seamuslee001 @eileenmcnaughton as it re-raises some the utf8mb4/extension questions.

Background
------------

The command `generate:entity-boilerplate` calls the SQL code-generator provided by core (ie `civicrm-core:CRM/CodeGen`). In core, the equivalent process reads [xml/schema/Schema.xml](https://github.com/civicrm/civicrm-core/blob/5.51.1/xml/schema/Schema.xml) which defines some general options/defaults for SQL tables, eg

```xml
<database xmlns:xi="http://www.w3.org/2001/XInclude">
<name>civicrm</name>
<character_set>utf8mb4</character_set>
<collate>utf8mb4_unicode_ci</collate>
<table_type>InnoDB</table_type>
<comment>Schema for CiviCRM</comment>

<xi:include href="Core/files.xml"              parse="xml" />
<xi:include href="ACL/files.xml"               parse="xml" />
<xi:include href="Contact/files.xml"           parse="xml" />
...
</database>
```

However, `civix` extensions do not have an analogous file. Instead, `civix` directly loads each of the table XML files. (*Original reason, IIRC -- It would be slightly cumbersome to maintain the top-level `Schema.xml` as it would require adding/removing various `xi:include` directives.*)

Before
------

The warnings from #249 indicate that these general options cannot be read.

(Specifically, `$config->database` should contain parsed version of the above; in actuality, `$config->database` is blank.)

After
-----

The warnings go away. `$config->database` is populated with some "reasonable" defaults that match the current policy in `civicrm-core`.

Comments
-----------

The current patch essentially hardcodes these options for purposes of `civix`-generated extensions. This raises a couple subquestions:

* Are we we OK with these defaults? These defaults produce slightly different output than before. Here's an example of how the output changes:

    ```diff
    [bknix-edge:~/bknix/build/wpmaster/web/wp-content/plugins/civicrm/civicrm/ext/myent] git diff sql/
    diff --git a/sql/auto_install.sql b/sql/auto_install.sql
    index a6e72bc..9c5f0c1 100644
    --- a/sql/auto_install.sql
    +++ b/sql/auto_install.sql
    @@ -39,4 +39,4 @@ CREATE TABLE `civicrm_foo` (
       PRIMARY KEY (`id`),
       CONSTRAINT FK_civicrm_foo_contact_id FOREIGN KEY (`contact_id`) REFERENCES `civicrm_contact`(`id`) ON DELETE CASCADE
     )
    -ENGINE=InnoDB;
    +ENGINE=InnoDB DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci ROW_FORMAT=DYNAMIC;
    ```
* How important is it to facilitate migration for downstream extensions to these defaults? (eg Do ext's need `upgrade_NNNN()` methods to transition existing deployments?)
* How important is it to allow extensions to customize the options? (Perhaps we should read these options from `info.xml`'s `<civix>` block or from `{$ext}/xml/schema/Schema.xml`.)

### DIFF
--- a/src/CRM/CivixBundle/Command/AddEntityBoilerplateCommand.php
+++ b/src/CRM/CivixBundle/Command/AddEntityBoilerplateCommand.php
@@ -71,8 +71,7 @@ class AddEntityBoilerplateCommand extends AbstractCommand {
     $config = new \stdClass();
     $config->phpCodePath = $basedir->string('');
     $config->sqlCodePath = $basedir->string('sql/');
-
-    $database['tableAttributes_modern'] = 'ENGINE=InnoDB';
+    $config->database = $this->getDefaultDatabase();
 
     foreach ($xmlSchemas as $xmlSchema) {
       $dom = new \DomDocument();
@@ -83,7 +82,8 @@ class AddEntityBoilerplateCommand extends AbstractCommand {
         $output->writeln('<error>There is an error in the XML for ' . $xmlSchema . '</error>');
         continue;
       }
-      $specification->getTable($xml, $database, $tables);
+      /** @var array $tables */
+      $specification->getTable($xml, $config->database, $tables);
       $name = (string) $xml->name;
       $tables[$name]['name'] = $name;
       $sourcePath = strstr($xmlSchema, "/xml/schema/{$ctx['namespace']}/");
@@ -200,6 +200,30 @@ class AddEntityBoilerplateCommand extends AbstractCommand {
         }
       }
     }
+  }
+
+  /**
+   * Get general/default database options (eg character set, collation).
+   *
+   * In civicrm-core, the `database` definition comes from
+   * `xml/schema/Schema.xml` and `$spec->getDatabase($dbXml)`.
+   *
+   * Civix uses different defaults. Explanations are inlined below.
+   *
+   * @return array
+   */
+  private function getDefaultDatabase(): array {
+    return [
+      'name' => '',
+      'attributes' => '',
+      'tableAttributes_modern' => 'ENGINE=InnoDB',
+      'tableAttributes_simple' => 'ENGINE=InnoDB',
+      // ^^ Set very limited defaults.
+      // Existing deployments may be inconsistent with respect to charsets and collations, and
+      // it's hard to attune with static code. This represents a compromise (until we can
+      // rework the process in a way that clearly addresses the inconsistencies among deployments).
+      'comment' => '',
+    ];
   }
 
 }


eileenmcnaughton:
In the past we specifically removed the Character set, collation line from extensions in order to allow them to pick up the char set from the database. That seems a bit moot now - ie I think sites should be pretty much transitioned to utf8mb4 now and I *think* the row format is the only one we would find in use . 

If the question is 'do extensions need an upgrade to adapt the char set. I would say it isn't necessary - it's arguably better not to specify the charset but I don't see much in it either way as they will pick it up from the DB.

- How important is it to allow extensions to customize the options? - as in charset etc - for the forseeable future there is no real need IMHO.

MegaphoneJon:
I would caution against assuming sites have transitioned to utf8mb4 simply because we never forced the change.  I only migrated most of my sites about 2 months ago, and I was aware of the need.  Lots of admins don't know this is necessary, and we should have a Civi upgrade force a change to utf8mb4 before we assume it's the standard.

demeritcowboy:
Anectodally I'd agree with @MegaphoneJon about whether or not sites have transitioned - I know a couple that haven't or have only done so recently. But additionally, converting civi tables to use utf8mb4 does not by itself change the database or system default in mysql. There are still some mysql hosts out there with latin1 (and I may have even seen myisam engine somewhere) as the system defaults.

Perhaps `getDefaultDatabaseXml()` could call `CRM_Core_BAO_SchemaHandler::getInUseCollation()` to figure out what to put. I've been using this in a few places:

```php
      $values['collation'] = CRM_Core_BAO_SchemaHandler::getInUseCollation();
      if (stripos($values['collation'], 'utf8mb4') !== FALSE) {
        $values['charset'] = 'utf8mb4';
      }
```

mattwire:
@totten Not sure about this. In my experience specifying the `Engine=InnoDB` is good/necessary but the character set causes problems. Eg. we had a lot of hardcoded `utf8` SQL in extensions etc. which overrides the default on the database even if the default is `utf8mb4`. Also if someone decides for some reason that they want to use something other than `utf8mb4_unicode` then this change breaks that. By the way I've seen problems in the past because drupal defaults to `utf8mb4_general` and CiviCRM defaults to `utf8mb4_unicode` and some queries get upset with the mixture. Obviously that's really a sysadmin problem but hardcoding it in an SQL statement makes it more difficult to resolve.

totten:
> (@eileenmcnaughton) In the past we specifically removed the Character set, collation line from extensions in order to allow them to pick up the char set from the database. 

Aah, right! This was the compromise that would allow *some possibility* for extensions to adapt to local collation circumstance, but it left some holes - and the longer-term goal was to generate SQL dynamically.

>  (@eileenmcnaughton)  "How important is it to allow extensions to customize the options?" - as in charset etc - for the forseeable future there is no real need IMHO.

Agree. It would take a complex situation and ... add more complexity... for no real need...

> (@MegaphoneJon) I would caution against assuming sites have transitioned to utf8mb4 simply because we never forced the change.

Agree. I looked around for migration resources and found (1) a status-check which suggests that Civi may (in the future) require utf8mb4 support in `mysqld` and (2) an API for performing a conversion. I couldn't find any mechanism (upgrade-tasks, upgrade-messages, status-checks) to get people to run `System.utf8conversion`.

> (@demeritcowboy) Perhaps `getDefaultDatabaseXml()` could call `CRM_Core_BAO_SchemaHandler::getInUseCollation()` to figure out what to put.

Thanks, `getInUseCollation()` looks handy. For `getDefaultDatabaseXml()` specifically, the timing is tricky -- `generate:entity-boilerplate` generates static SQL at dev-time. Adapting to each local deployment (via `getInUseCollation()`) would require running at install-time.

I guess one way to adapt at install-time would be to set some Smarty variables (`COLLATE {$schema.collation}`) and rename `auto_install.sql` to `auto_install.mysql.tpl`? (That's a bigger change -- this PR is really just trying to fix the PHP warning.)

> (@mattwire)  In my experience specifying the `Engine=InnoDB` is good/necessary but the character set causes problems... hardcoded `utf8` SQL in extensions  ... drupal defaults ... some queries get upset with the mixture ...

There's an interesting distinction in there. We can take further steps toward normalizing civicrm-core and extensions (eg discourage hardcoded `utf8` with blogs, code-warnings, universe-audits). But for shared databases (Civi+CMS) and integrations like `views`, you can get painted into a corner. I guess the question is whether to care about that corner? If we need to support that kind of thing indefinitely, then standardizing on `utf8mb4` wouldn't be viable -- we'd need to all-in on dynamic SQL.

Long-term, I like that standardizing `utf8mb4` leads to an broadly simpler arrangement, and I think Drupal `views`-style integration will probably fade away (due to `civicrm_entity`, `search_kit`, etc). But shared databases will continue indefinitely, and (*for near/mid-term - where the overall situation is muddled*) dynamic SQL would probably be more forgiving/interoperable/compatible.

Anyway, we've drifted far from the warnings in #249. For that... I'm pushing an update to set `$config->database` in a different way. It should solve the warnings and give the same output as the status-quo (ie just `ENGINE=InnoDB`). We can+should open other issues/PRs for resolving utf8mb4.

demeritcowboy:
Noting there's at least one lab ticket: https://lab.civicrm.org/dev/core/-/issues/2237. There was general discussion here https://lab.civicrm.org/dev/core/-/issues/339 including extension table creation, but I can't find one specifically on the topic.

PR #258, opened 2022-08-17
Add some (easy) return type hints

eileenmcnaughton:
php is getting tighter & tigher on type hints & I think it's pretty likely that by the
time we are doing php9 they we will be doing re-work to add them - so
we should make sure we are adding them on code as we go now. This just adds
a handful of easy ones (ones where it is really easy to see what the hints
should be)


totten:
Agree these type-hints look sensible. The tests seem to work fine with the patch. Merging.

PR #259, opened 2022-08-17
Use fruit as examples instead of a wanker

agileware-justin:
Use fruit as examples instead of a wanker which is a bit more PC.


totten:
They say Apples are better than PCs...

agileware-justin:
Only if they don't have the skillz to run Linux :penguin:

PR #260, opened 2022-08-24
Fix self-reference being identified as circular foreign key

pfigel:
This fixes an issue where foreign keys referencing their own table trigger the check for circular dependencies. This can be avoided by ignoring affected foreign keys when ordering tables.

This also fixes a null reference in the error handler.

Note that I've only done limited testing (read: running `tests/e2e/AddEntityTest.php`) as I ran into an issue with `box` and ran out of time to work on this. 😅


totten:
Nice catch @pfigel. Test update looks good (and runs cleanly for me). Merging.

PR #265, opened 2022-10-07
Add PSR-0 classloading for "CRM"

totten:
This updates the `info.xml` in new extensions (`civix generate:module`) and existing extensions (`civix upgrade`) to declare:

```xml
<classloader>
  <psr0 prefix="CRM_" path=""/>
</classloader>
```

Todo: Check test results

### DIFF
--- a/src/CRM/CivixBundle/Builder/Info.php
+++ b/src/CRM/CivixBundle/Builder/Info.php
@@ -187,4 +187,45 @@ class Info extends XML {
     return empty($mixins) ? '13.10.0' : '22.05.0';
   }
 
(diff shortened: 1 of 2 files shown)


demeritcowboy:
When I had [looked at this before](https://github.com/civicrm/civicrm-core/pull/24580#issuecomment-1256415695) I needed to make path "." - the empty string didn't work (I don't remember the reason).

totten:
Aah, thanks. The `"."` seems to work in the training scenario that led to this PR. I'm not certain the difference either, but if you've seen cases where `"."` is better, then let's do that.

Pushed update.

PR #270, opened 2022-11-12
Upgrade to use CRM_Extension_Upgrader_Base and info.xml `<upgrader>` tag

colemanw:
Since 5.38 core supports an `<upgrader>` tag in liu of hooks, and a common base class.
For extensions where core version is >= 5.38, this will generate upgrader classes that use the common base, and the tag instead of delegated hooks.
It will also perform this cleanup on existing modules during each `civix upgrade`.

### DIFF
--- a/src/CRM/CivixBundle/Builder/Info.php
+++ b/src/CRM/CivixBundle/Builder/Info.php
@@ -11,6 +11,11 @@ use SimpleXMLElement;
 class Info extends XML {
 
   public function init(&$ctx) {
+    $ctx += [
+      // FIXME: Auto-detect current installed civi version
+      'compatibilityVerMin' => 5.45,
+    ];
+
     $xml = new SimpleXMLElement('<extension></extension>');
     $xml->addAttribute('key', $ctx['fullName']);
     $xml->addAttribute('type', $ctx['type']);
@@ -40,7 +45,7 @@ class Info extends XML {
     $xml->addChild('releaseDate', date('Y-m-d'));
     $xml->addChild('version', '1.0');
     $xml->addChild('develStage', 'alpha');
-    $xml->addChild('compatibility')->addChild('ver', $ctx['compatibilityVerMin'] ?? '5.0');
+    $xml->addChild('compatibility')->addChild('ver', $ctx['compatibilityVerMin']);
     $xml->addChild('comments', 'This is a new, undeveloped module');
 
     // APIv4 will look for classes+files matching 'Civi/Api4', and
--- a/src/CRM/CivixBundle/Command/AddUpgraderCommand.php
+++ b/src/CRM/CivixBundle/Command/AddUpgraderCommand.php
@@ -37,12 +37,21 @@ class AddUpgraderCommand extends AbstractCommand {
       return;
     }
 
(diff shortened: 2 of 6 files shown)


colemanw:
@totten This uses the general upgrader rather than a specific upgrade version (as the `<upgrader>` tag depends on core version, not civix version). People on the latest civix version would need to manually trigger an upgrade... or we could add an empty upgrade step.

On a related note, I think we should consider adding an upgrade step that prompts the user to bump min core version to whatever ESR is set at. This could potentially be dynamic so we don't have to periodically patch civix whenever a new esr comes out.

colemanw:
@totten now that I think about it, I think creating an empty upgrade step will be necessary because `civix.php` files get rewritten more often than just during upgrades, and we need to make sure the delegated hooks get removed in tandem.

I still think prompting the user to bump their core version would be really nice. Otherwise civix doesn't give them any new features.

totten:
General reaction: woot!

For `r-run`, I took a copy of of `mosaico` and tried the following:

1. Install an old version (2.4). Switch to a newer version (2.10) and run the old-style upgrader. Observe that some of the schema evolves (e.g. `civicrm_mosaico_template` gets new `domain_id`).
2. Install an old version (2.4). Switch to a newer version (2.x) and apply civix updates. Run the new-style upgrader. Observe that some of the schema evolves in the same way (e.g. `civicrm_mosaico_template` gets new `domain_id`).

So that looked good.

I locally rebased to current `master` and ran the tests. The last commit (["Bump default civi version to 5.45"](https://github.com/totten/civix/pull/270/commits/e1b9ff58d3ef054aed9cf642ef109512a729df7e)) caused a bunch of test failures. So I reverted that and moved that topic to a separate PR (#273).

totten:
Erg, ran into a problem. Not sure why I missed it before,  but the `make-example.sh` is no longer running properly for me. The issue: each stub had two parts (in `module.civix.php` and `module.php`). The PR removed the `_install` stub from `module.civix.php`, but (for new extensions) it still creates the part for `module.php`.

After re-reading, I *think* there's a second problem. It's correct that `_install` is no longer needed for loading the `CRM_*_Upgrader`. But `_install` has a second job -- loading the mixin-polyfill (`$_invokePolyfill`). I'm pretty sure this was added in response to some kind of bug (*e.g. you enable a new extension, but the polyfill doesn't activate quickly enough, so your mixins aren't used until a subsequent pageload; it's hard to say exactly because the install process does a lot of flushing*). The polyfill is used on Civi<5.45.

The simplest thing would be to partially restore the stubs -- but only for purposes of `$_invokePolyfill` (unless we can come up with a better idea in the next day or so).

colemanw:
`<unhelpful>`
You know, a lot of headaches could be saved if we didn't have to delegate hooks and a `.civix.php` file could implement hooks directly. If only something like `function myextension_civicrm_install_2()` was supported. Kinda late for that since we're migrating towards events (which do allow the same extension multiple callbacks) but it woulda been nice 10 years ago...
`</unhelpful>`

Yea I think it's fine to partially restore the delegated functions but only for <5.45 & remove them completely in modern versions.

PR #271, opened 2022-11-12
Use 2nd argument of basename instead of preg_replace

colemanw:
Sorta like https://github.com/civicrm/civicrm-core/pull/24948 but for civix


totten:
Looks good, and tests pass.

PR #272, opened 2022-11-13
Use entity-type-php mixin

colemanw:
Adds an upgrader for the new mixin. Depends on https://github.com/civicrm/civicrm-core/pull/24947


colemanw:
@totten I'm still unclear on why you think this mixin might not work. See my question here: https://github.com/civicrm/civicrm-core/pull/24947#issuecomment-1311961298

totten:
(1) Noted in the other core thread -- in my testing,the mixin seemed to work with `core>=5.45` but not `core<=5.44`. That covers two ESR releases (5.51 and 5.45), so I think it's fair for entity-generation to require `5.45`.

(2) There needs to be a registration record along the lines of this (*the snippet I used for testing last night*):

```diff
diff --git a/mixin-backports.php b/mixin-backports.php
index b2ebdf0..918943d 100644
--- a/mixin-backports.php
+++ b/mixin-backports.php
@@ -35,6 +35,14 @@ return [
     'provided-by' => '5.45.beta1',
     'minimum' => '5.27', /* Compat may go back further; haven't tested */
   ],
+  'entity-types-php@1' => [
+    'version' => '1.0.0',
+    'sha256' => 'f8e10aac991b2b3acac269a1fca81f883a295908990687959e790885a2e410c2',
+    'remote' => 'https://raw.githubusercontent.com/civicrm/civicrm-core/8de7881b35e3f98d38b147dcf23f5a44d5dfe035/mixin/entity-types-php%401/mixin.php',
+    'local' => 'extern/mixin/entity-types-php@1/mixin.php',
+    'provided-by' => '5.57.beta1',
+    'minimum' => '5.27', /* Compat may go back further; haven't tested */
+  ],
   'menu-xml@1' => [
     'version' => '1.0.0',
     'sha256' => '4f5be44d6764816b22d0a5cdc2e047cfd9ec4acf48e548f82bb20c05db933d0e',
```

(That obviously needs some tweaking - e.g. bump `minimum` to `5.45` and maybe get a prettier `remote` URL.)

(3) So... how to put some force to the `minimum`.... the smallest patch might update `\CRM\CivixBundle\Builder\Mixins::addMixinToXml()` so that it revises both `<mixins>` and `<compatibility>`. The change would become apparent in git (`git diff` / `git commit`), though it would be more polite to sprinkle in some confirmations. (e.g. `22.12.0.up.php` would warn that it's going to raise the `<compatibility>`).

colemanw:
@totten that all sounds good to me. So to recap, if I run `civix generate:entity` it will check core compatibility and confirm if it needs a bump to 5.45, and abort if I type `n`. Effectively that means civix will not generate entities for modules with older core versions, which I agree is perfectly reasonable.

totten:
Rebased on top of current master and sibling PR #274.

tests/e2e/CivixProjectTestTrait.php
```diff
@@ -117,6 +117,22 @@ public function civixGenerateEntityBoilerplate(): CommandTester {
     return $tester;
   }
 
+  /**
+   * Get a value from "info.xml" by calling "civix info:set"
```
colemanw, line 121:
```suggestion
   * Get a value from "info.xml" by calling "civix info:get"
```

mixin-backports.php
```diff
@@ -35,6 +35,14 @@
     'provided-by' => '5.45.beta1',
     'minimum' => '5.27', /* Compat may go back further; haven't tested */
   ],
+  'entity-types-php@1' => [
+    'version' => '1.0.0',
+    'sha256' => 'f8e10aac991b2b3acac269a1fca81f883a295908990687959e790885a2e410c2',
+    'remote' => 'https://raw.githubusercontent.com/civicrm/civicrm-core/8de7881b35e3f98d38b147dcf23f5a44d5dfe035/mixin/entity-types-php%401/mixin.php',
```
colemanw, line 41: @totten so this url will be prettier once https://github.com/civicrm/civicrm-core/pull/24947 is merged right? IMO let's go ahead and merge it.

colemanw: Ok merged. This URL can be updated.

totten:
Rebased and updated some of the numbering. 

CI test run looks good in a recent run (very-similar commit) with `core`@`master`,`5.57`,`5.51`. So expecting the next run to look good.

Locally, I've also tested with `core`@`5.45` because that's the version we marked before as the min-compat for `entity-type-php`.  And it was passing.

We should merge after he next test-run passes.

PR #273, opened 2022-11-22
"generate:module" - Add option to specify compatibility of new modules. Use dynamic default.

totten:
This does three related this:

1. Allow an option like `civix generate:module --compatibility=5.0`
2. Use a default value (`--compatibility=current`) which inherits the current version from the locally installed CiviCRM instance. (Presumably, that's the version you will actually test against...)
3. Update some of the E2E tests to explicitly control for the initial `--compatibility`. These tests examine how the code changes with different compatibility targets.


colemanw:
This looks good @totten.
+1 for merging.

PR #275, opened 2022-11-26
Drop unused "codecept" integration

totten:
I never used this, and I can't find anyone in `universe` who has published code with it.  It adds makework/dependencies/weight for QA of civix, and there's no point.

(If someone did want to follow-through on an intitiative for codecept, then...  I think that would be a de-facto new initiative.  Maybe this patch would be a useful reference...  but that doesn't mean we need to maintain it speculatively...)


demeritcowboy:
Can we also drop it from https://github.com/totten/civix/blob/a3bb6049a02268608fa90068446d531a2417860e/src/CRM/CivixBundle/Resources/views/Code/phpunit-boot-cv.php.php#L38 ? And then I can stop needing to do the hack in https://github.com/totten/civix/pull/153/files. In fact I always just comment out that line since it doesn't seem needed.

PR #277, opened 2022-11-29
Reinstante lifecycle hooks (partially). Always uses 5.38-style `<upgrader>`.

totten:
This is a follow-up/re-roll to the recent #270. Its main purpose is to fix the regression where new extensions have mis-linked stubs. The approach is to go all-in on 5.38-style `<upgrader>` classes -- but to preserve the edifice needed for mixin-polyfill.

## Example

If you have an extension that targets `5.0` and defines `CRM_*_Upgrader`, then `civix upgrade` will produce this:

Screen Shot 2022-11-28 at 6 14 13 PM

The message gets simpler in other circumstances -- e.g.

* If you don't have `CRM_*_Upgrader`, then some of the bullets go away.
* If you don't have `CRM_*_Upgrader`, or if you already require 5.38+, then the warning goes away.

## Technical Details

* `CRM_*_Upgrader`
    * (**Invariant condition for v22.12**) `CRM_*_Upgrader` requires CiviCRM v5.38.
    * If you have an extension which already uses `CRM_*_Upgrader`, then `civix upgrade` will update `info.xml` (ie `<upgrader>` and possibly `<compatibility>`).
    * If you have a new extension and call `civix generate:upgrader`, then it will update `info.xml` (ie `<upgrader>` and possibly `<compatibility>`).
* `hook_postInstall`, `hook_disable`, `hook_uninstall`
    * (**Invariant condition for v22.12**) civix does *not* use these hooks.
    * (`mymodule.civix.php`) Removed these hook implementations (across-the-board/unconditionally)
    * (`mymodule.php`) Removed stubs for new renditions of `mymodule.php`. For existing renditions of `mymodule.php`, perform a soft-removal (*ie drop the old delegation code; if the function becomes empty, then drop that*).
* `hook_install` and `hook_enable`
    * (**Invariant condition for v22.12**) civix does use these these hooks.
    * (`mymodule.civix.php`) Preserve these hook implementations. However, the content becomes shorter/simpler -- they no longer call `CRM_*_Upgrader`.
    * (`mymodule.php`) Preserve these stubs.
    * (**Note**) As always, clever developers who know what they're doing can delete stubs from `mymodule.php`.
    * (**Note**) We could revisit after (say...) March 2023 and consider raising the minimum to 5.45 -- which should make these unnecessary.

## Edge Case: hook_install and "include paths"

~~There is one edge-case that may be broken by this regime. (This probably broke in #270, but I reinforced the breakage in this branch.) The old versions of `hook_install` (etal) delegated down to to `_civix_civicrm_config()` to ensure that the include-paths were setup. Historically, this was important for class-loading - but it shouldn't matter now that v22.10+ defines `<psr0>`.. It could theoretically be an issue for Smarty's include-path -- but I've searched `universe` and only found two places where an upgrader references `smarty` or `tpl`. This PR checks for similar references and prompts the developer to investigate:~~

Screen Shot 2022-11-28 at 6 47 43 PM

(*REVISED*) Rereading this, it strikes me that the include-path issue also affects APIv3. It is much more plausible for that cuase trouble. Need to look at that.

(*There's also potential impact for class-overrides, but it's extremely edgy, and my interest in investigating that angle is... ehm... *)

### DIFF
--- a/scripts/load-snapshot.sh
+++ b/scripts/load-snapshot.sh
@@ -1,0 +1,107 @@
+#!/usr/bin/env bash
+
+## Load an snapshot, i.e.
+## - Clean the test site.
+## - Extract the snapshot into the test site.
+## - Install the extension
+
+################################################
+function usage() {
+  echo "$0 [-r] [-x] [-g] [-e] [-u] [-s <source>] "
+  echo
+  echo "Data:"
+  echo "-s     (Source)   Specify which snapshot file to load"
+  echo
+  echo "Actions:"
+  echo "-r     (Reset)    Reset Civi DB"
+  echo "-x     (Extract)  Extract the zip file to a temporary folder"
+  echo "-g     (Git)      Add temporary folder to a temporary git repo"
+  echo "-e     (Enable)   Enable the extension in CiviCRM"
+  echo "-u     (Upgrade)   Run civix upgrades"
+  echo
+  echo "Example:"
+  echo "  $0 -rxge -s tests/snapshots/org.example.civixsnapshot-v19.11.0-kitchensink/"
+  echo
+  echo "NOTE: Actions are executed in requested order. So these two are slightly different:"
+  echo "  $0 -rxgeu    (restore => extract => git => enable => upgrade)"
+  echo "  $0 -rxgue    (restore => extract => git => upgrade => enable)"
+}
+
+################################################
+EXMODULE=${EXMODULE:-org.example.civixsnapshot}
+#SNAPSHOT="$1"
+#shift
+SNAPSHOT=
+TASKS=
+while getopts "hs:rxgeu" opt; do
+  case $opt in
+    h) usage ; exit ; ;;
+    s) SNAPSHOT="$OPTARG" ; ;;
+    r) TASKS="$TASKS restore" ; ;;
+    x) TASKS="$TASKS extract" ; ;;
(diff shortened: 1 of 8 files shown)


totten:
Reinstated include-path for hook_enable/hook_install. 

`civix/tests/e2e/**php` seems to run clean on `core` with `5.38`,`5.51`, and `master`.

There are still some fiddly bits, but this is better in-than-out. Of note:

* I haven't properly tried out scenarios that rely on include-path during install/etc.
* I'm getting some trouble with running generated-tests on 5.51. (This could just reflect a lack of testing before -- not necessarily caused by this PR.) 

```
./scripts/make-snapshots.sh --phar kitchensink --test --keep
```

```
Installing dmastertest_qmzfj schema

Fatal error: Uncaught Error: Class 'Civi\Api4\SearchSegment' not found in /Users/totten/bknix/build/dmaster/web/sites/all/modules/civicrm/ext/search_kit/Civi/Api4/Service/Spec/Provider/SearchSegmentExtraFieldProvider.php on line 52

Error: Class 'Civi\Api4\SearchSegment' not found in /Users/totten/bknix/build/dmaster/web/sites/all/modules/civicrm/ext/search_kit/Civi/Api4/Service/Spec/Provider/SearchSegmentExtraFieldProvider.php on line 52
```

PR #285, opened 2023-01-16
Only add templates dir if it exists

eileenmcnaughton:
On trying to upgrade some extensions `addTemplateDir` actions I was struck by how many of the extensions are adding template directories that don't exist. I feel like this must be at least marginally worse for performance & it feels a bit confusing

@totten


totten:
Test failures are unrelated.

I haven't benchmarked, but I agree with the intuition - the status-quo is probably marginally worse for performance. From the POV of just prepending one extension that does not have `templates_c`':

* If you add the path and then include it in searches, then it's one `stat()` per active `tpl` per page-load. 
* If you do the file-exists check here, it's one `stat()` per page-load.

PR #300, opened 2023-07-12
Update cv library. Respect CIVICRM_BOOT or CIVICRM_SETTINGS.

totten:
This update aims to resolve an issue identified in the discussion of https://github.com/civicrm/cv/pull/166 - i.e. `civix` doesn't support some of bootstrap (initialization) options that `cv` does, which limits compatibility with some environments.

Before
---------

* When `civix` needs to acces CiviCRM, it uses `cv`'s `Bootstrap.php`.
* `Bootstrap.php` is included by way of a special/single-file download step.
* `Bootstrap.php` only supports the traditional (Civi-first / CIVICRM_SETTINGS) protocol. It doesn't support the newer (CMS-first / CIVICRM_BOOT) protocol.

After
-------

* When `civix` needs to acces CiviCRM, it uses `cv`'s `Bootstrap.php` _or_ `CmsBootstrap.php`.
* All `cv` helper classes are included as a composer dependency (`civicrm/cv-lib`).
* `civix`'s boot behavior should now match `cv`'s behavior, as described in https://github.com/civicrm/cv#bootstrap

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,6 +5,7 @@
     "license": "AGPL-3.0",
     "require": {
         "php": ">=7.3.0",
+        "civicrm/cv-lib": "~0.3.44",
         "civicrm/composer-compile-plugin": "~0.18",
         "civicrm/composer-downloads-plugin": "~2.1|^3",
         "symfony/console": "^4|^5",
@@ -19,7 +20,6 @@
     "autoload": {
         "psr-0": {
             "Mixlib": "extern/src/",
-            "Civi\\Cv": "extern/src/",
             "CRM\\CivixBundle": "src/"
          }
     },
@@ -50,12 +50,6 @@
             }
         ],
         "downloads": {
-          "bootstrap": {
-              "version": "v0.3.43",
-              "url": "https://raw.githubusercontent.com/civicrm/cv/{$version}/src/Bootstrap.php",
-              "path": "extern/src/Civi/Cv/Bootstrap.php",
-              "type": "file"
-          },
           "mixlib": {
               "version": "5.45",
               "url": "https://raw.githubusercontent.com/civicrm/civicrm-core/{$version}/tools/mixin/src/Mixlib.php",
--- a/composer.json
+++ b/composer.json
@@ -4,7 +4,7 @@
     "homepage": "https://github.com/totten/civix",
     "license": "AGPL-3.0",
     "require": {
-        "php": ">=7.2.0",
+        "php": ">=7.3.0",
         "civicrm/composer-compile-plugin": "~0.18",
         "civicrm/composer-downloads-plugin": "~2.1|^3",
         "symfony/console": "^4|^5",
@@ -32,7 +32,7 @@
           "civicrm/composer-downloads-plugin": true
         },
         "platform": {
-          "php": "7.2.0"
+          "php": "7.3.0"
         },
         "bin-dir": "bin"
     },
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "6d25be2fb982e1b021998a7e2fba372b",
+    "content-hash": "51b24f21ca09ac6dbc30b84efe25fc77",
     "packages": [
         {
             "name": "civicrm/composer-compile-plugin",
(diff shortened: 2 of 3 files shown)


jensschuppe:
Thanks, @totten, that's great work! Unfortunately, there's now a Composer dependency mismatch causing a

`PHP Fatal error:  Declaration of Drupal\dblog\Logger\DbLog::emergency(Stringable|string $message, array $context = []): void must be compatible with Psr\Log\LoggerInterface::emergency($message, array $context = []) in /path/to/drupal/web/core/lib/Drupal/Core/Logger/RfcLoggerTrait.php on line 20`.

For civix:
```
$ composer depends psr/log
civicrm/cv-lib  v0.3.44 requires  psr/log (~1.1 || ~2.0 || ~3.0) 
symfony/console v4.4.34 conflicts psr/log (>=3)
```
while the Drupal Composer project has `psr/log:3.0.0` (but at least `2.0.0` is compatible).

```
$ composer why-not psr/log 2.0.0
psr/log 2.0.0 requires php (>=8.0.0 but 7.3.0 is installed)
```

*Civix* uses the older version due to its `config.platform` setting for `php` being set to `7.3.0`. Is that on purpose?

With *civix* not locking the PHP version via `config.platform`, this works well.

totten:
> Civix uses the older version due to its `config.platform` setting for `php` being set to `7.3.0`. Is that on purpose?

Yup. It matches the declarations on `civicrm-core` and `cv`. The aim is to work anywhere that CiviCRM does.

The `config.platform` ensures that we do our testing/QA/release with versions that are suitable for use in the PHARs. Without it, we'd quickly get packages that deviate from the minimum requirements.

(Key difference - PHARs must have packages that are compatible with the full range of environments, i.e. php73-php82. But when you locally run `composer` on `civix.git`, it only needs a version that's compatible with the local system.) 

> With civix not locking the PHP version via `config.platform`, this works well.  Installing _Composer_ dependencies for _civix_ with `--ignore-platform-reqs` is a workaround. I'm not sure the packaged _civix_ phar does this or not, as I'm using the repo.

I haven't yet reproduced the problem-scenario -- I tried a bit with D9. What's your UF? D10?

In *principle*, the `LoggerInterface` problem shouldn't affect `civix.phar` because it uses [namespace prefixing](https://github.com/humbug/php-scoper). But *actually* there's a bug where namespace prefixing is de-facto off.  `cv` had the same issue a few months back. I'm expect we can adapt https://github.com/civicrm/cv/pull/158 for a similar fix on `civix`.

But parallel to that, `psr/log` has become a bit of a hassle since they started bumping major-versions. I dunno -`cv-lib` is a pretty early/low-level thing. Maybe it use an internal `interface` instead of `psr/log`...

demeritcowboy:
I'm seeing that error with drupal 10. php is 8.1 (drupal 10 won't run with less). psr/log is 3.0.0. Happens during core:install.

`PHP Fatal error:  Declaration of Drupal\dblog\Logger\DbLog::emergency(Stringable|string $message, array $context = []): void must be compatible with Psr\Log\LoggerInterface::emergency($message, array $context = []) in .../web/core/lib/Drupal/Core/Logger/RfcLoggerTrait.php on line 20`

Since it sounds like the problem is that cv is getting built with an older psr/log, and in my case it's a build of cv from git, I can probably just tell it to use 3.0.

jensschuppe:
> > With civix not locking the PHP version via config.platform, this works well. Installing Composer dependencies for civix with --ignore-platform-reqs is a workaround. I'm not sure the packaged civix phar does this or not, as I'm using the repo.
> 
> I haven't yet reproduced the problem-scenario -- I tried a bit with D9. What's your UF? D10?

Drupal 10 with PHP 8.1.

totten:
Dave [mentioned some success](https://github.com/civicrm/cv/pull/169#issuecomment-1634174450) after dropping `psr/log` requirement.

I've been trying to port https://github.com/civicrm/cv/pull/158 (but it's not working yet)

totten:
@demeritcowboy Yeah, I wasn't aware of that specific conflict, but that's the kind of thing that namespace-prefixing should address. (i.e. There's nothing innately broken with Symfony4, except that Symfony4 and Symfony6 don't naturally coexist.)

Anyway, I made some progress - and this PR produces a PHAR that looks plausibly prefixed. I' can do some basic commands on my local D10+php81. (However, I'm not sure my local environment is hostile enough for testing it.) 

* https://github.com/totten/civix/pull/302

totten:
Well, those bits are merged (*dropping the `psr/log` requirement and re-enabling the namespace-prefixes on `civix.phar`*). As far as my quick/manual inspection goes, they seem to work...

So then I tried to setup the test-matrix to include D9+D10. This is where it's a bit ucky.

* [D9 is failing](https://test.civicrm.org/job/CiviCRM-Civix-Matrix/BKPROF=max,BLDTYPE=drupal9-clean,CIVIVER=master,label=bknix-tmp/lastCompletedBuild/testReport/) on several tests with failures like this:

    ```
    TypeError: htmlentities(): Argument #1 ($string) must be of type string, array given
    
    /home/homer/buildkit/build/build-2/vendor/civicrm/civicrm-core/CRM/Utils/System.php:284
    /home/homer/buildkit/build/build-2/vendor/civicrm/civicrm-core/CRM/Core/Menu.php:445
    /home/homer/buildkit/build/build-2/vendor/civicrm/civicrm-core/CRM/Core/Menu.php:265
    /home/homer/buildkit/build/build-2/vendor/civicrm/civicrm-core/CRM/Core/Menu.php:296
    /home/homer/buildkit/build/build-2/vendor/civicrm/civicrm-core/CRM/Core/Invoke.php:389
    /home/homer/buildkit/build/build-2/vendor/civicrm/civicrm-core/CRM/Extension/Manager.php:319
    ```

* [D10 is failing](https://test.civicrm.org/job/CiviCRM-Civix-Matrix/BKPROF=max,BLDTYPE=drupal10-clean,CIVIVER=master,label=bknix-tmp/30/console) with a Symfony class-contract error.

    ```
    ...
    Test 'E2E\MultiExtensionTest::testAddPage' started
    Test 'E2E\MultiExtensionTest::testAddPage' ended
    ...
    Test 'E2E\SnapshotUpgradeTest::testSnapshot with data set "civixsnapshot-HEAD-empty" ('org.example.civixsnapshot-HEAD-empty')' started

    Fatal error: Declaration of Symfony\Component\Console\Style\OutputStyle::write(iterable|string $messages, bool $newline = false, int $type = self::OUTPUT_NORMAL) must be compatible with Symfony\Component\Console\Output\OutputInterface::write($messages, $newline = false, $options = 0) in /home/homer/buildkit/build/build-4/vendor/symfony/console/Style/OutputStyle.php on line 49
    ```

I haven't really looked into the D9 one. I tried the D10 first since it sounded thematically similar to the issue @demeritcowboy had before (and I was hoping to show the test passing after the merges). But the test+failure seem a bit gnarly:

1. The failure doesn't occur when running `SnapshotUpgradeTest` by itself. It requires some interaction with earlier tests.
2. I *suspect* the issue  is roughly like this: the test is based on Symfony `CommandTester`, which does everything in one process - and this process doesn't benefit from PHAR namespace-prefixing. So it invites a sort of race between two copies of Symfony.

The first solution that comes to mind is to rearrange this test to use `civix.phar`, which doesn't see easy. And the thing is that this is separate from substantive compatibility issues -- it's an issue in just the test harness.

demeritcowboy:
That first one I have come across before, and I remember just hacking it and then thinking I'd come back to it later. Let me see if I have any notes somewhere.

demeritcowboy:
What I came across was similar but different. I did come back to it and had a series of PRs, these being the closest, but are different:
* https://github.com/civicrm/civicrm-core/pull/24176
* https://github.com/civicrm/civicrm-core/pull/24192

So as you've noted for the drupal 10 part this might be specific to the e2e tests.

PR #301, opened 2023-07-13
composer.json - Update to civicrm/cv-lib v0.3.45

totten:
See discussion on https://github.com/totten/civix/pull/300


totten:
civibot, test this please

PR #302, opened 2023-07-13
php-scoper - Fix prefixing

totten:
The purpose of `php-scoper` is to allow `civix.phar` to run in other environments without provoking class-conflicts. It does this by prefixing (almost all) classes.

The current configuration appears to be inert -- the final output doesn't actually have prefixes nowadays. This brings it back and adds some test-coverage.

Comments
--------------

This is very similar to https://github.com/civicrm/cv/pull/158, except that I had to do some fine-tuning on the include/exclude lists.

### DIFF
--- a/box.json
+++ b/box.json
@@ -14,6 +14,7 @@
         {
             "name": "*.php",
             "exclude": [
+              "composer-compile-plugin",
               "phpunit",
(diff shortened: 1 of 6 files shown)


totten:
civibot, test this please

PR #303, opened 2023-07-17
Remove commented-out boilerplate

colemanw:
Neither of these commented-out functions sets a particularly good example so it ends up just being noise.


colemanw:
@totten

PR #304, opened 2023-07-23
generate:form gives PHP Fatal error:  Uncaught TypeError: Return value of "CRM\CivixBundle\Command\AddFormCommand::execute()" must be of the type int, "null" returned. in ...\vendor\symfony\console\Command\Command.php:301

demeritcowboy:
In this case I'm using php 8.1 and civix built from source.

It's very similar to https://lab.civicrm.org/dev/civicrm-asset-plugin/-/merge_requests/8/diffs in that more recent versions of symfony require execute() to return an int: https://github.com/symfony/console/blob/v5.4.5/Command/Command.php#L301. And older versions wouldn't care if it's 0 or null.

### DIFF
--- a/src/CRM/CivixBundle/Command/AbstractAddPageCommand.php
+++ b/src/CRM/CivixBundle/Command/AbstractAddPageCommand.php
@@ -92,6 +92,7 @@ abstract class AbstractAddPageCommand extends AbstractCommand {
     $mixins = new Mixins($info, $basedir->string('mixin'), ['menu-xml@1.0', 'smarty-v2@1.0']);
     $mixins->save($ctx, $output);
     $info->save($ctx, $output);
+    return 0;
   }
 
   abstract protected function getPhpTemplate(InputInterface $input);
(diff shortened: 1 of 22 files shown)


totten:
civibot, add to whitelist

totten:
@demeritcowboy Was there a reason for targeting this command specifically? It looked like other commands would have same issue. I pushed what I expect them to need.

Aside: If your environment can run it, it might be nice to see if this command works:

```
cd civix
export CIVIX_WORKSPACE=/path/to/sites/all/modules/civicrm/ext/civixtest 
bash scripts/make-snapshots.sh --src --keep kitchensink
```

That will create a "kitchensink" extension (*with a little bit of code from all the generators*). If any command still have a similar issue, then generating the `kitchensink` should fail.

demeritcowboy:
No it was just because that's what I was doing at the time.

That script wouldn't work for me but if it's just a matter of making a module and then running [this list](https://github.com/totten/civix/blob/fded98f6ae2a9c8ea5f271b3a0b91014c3ca64e0/scripts/make-snapshots.sh#L99C7-L121C50), then except for some warnings and the single quotes I had to change it all seems to work.

totten:
@demeritcowboy Yup, that's exactly the list. Glad it worked.

demeritcowboy:
Thanks!

PR #309, opened 2023-09-08
refs #composer updated Symfony components

kewljuice:


totten:
civibot, add to whitelist

colemanw:
Does the `composer.lock` file need to be updated as well?

totten:
Conflicted. 

Note that (in the intervening period) we have  updated `composer.lock` now to use v5.

PR #313, opened 2023-10-11
Superficial fixes to form template - hints, array syntax

eileenmcnaughton:
This function is pretty prevalent in civix generated stuff - could be a mixin or core function - but probably just getting it to nice arrays etc is a step  forwards

After running it the only php grumble is it can't see that the class is called - @totten would be cool if your phpstorm extension could pick up that xml declarations make it called - but I assume that's a big lift

image

### DIFF
--- a/src/CRM/CivixBundle/Resources/views/Code/form.php.php
+++ b/src/CRM/CivixBundle/Resources/views/Code/form.php.php
@@ -11,7 +11,11 @@ use <?php echo $_namespace ?>_ExtensionUtil as E;
  * @see https://docs.civicrm.org/dev/en/latest/framework/quickform/
  */
 class <?php echo preg_replace(':/:', '_', $fullClassName) ?> extends CRM_Core_Form {
-  public function buildQuickForm() {
+
+  /**
+   * @throws \CRM_Core_Exception
+   */
+  public function buildQuickForm(): void {
 
     // add form elements
     $this->add(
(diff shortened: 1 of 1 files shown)


totten:
Patch looks good. Test-run is happy. (I'm pretty any hard-to-see typos would cause a failure.) Let's merge.

> After running it the only php grumble is it can't see that the class is called - @totten would be cool if your phpstorm extension could pick up that xml declarations make it called - but I assume that's a big lift

* In the [phpstorm.meta.php spec](https://www.jetbrains.com/help/phpstorm/ide-advanced-metadata.html), I don't see an obvious way to express a "Find usage"-type-reference between an XML file and PHP file.
* It looks like Jetbrains plugin API could be used to [define references](https://plugins.jetbrains.com/docs/intellij/psi-references.html#contributed-references). But we'd have to write/maintain a real "plugin". (This is traditionally Java-based, but there might be [an option for JS-based micro-plugin](https://plugins.jetbrains.com/plugin/17669-flora-beta-).) That is a deeper dive.
* A quick hack might be to... generate a _plain PHP file_ that mirrors the routing table? The "Find Usages" wouldn't go directly to the relevant XML, but I guess it would point in the right direction.

```php
<?php
// AUTO-GENERATED: This mirrors the routing table so that the IDE gives some more hints about usage.

define('CORE', '/home/totten/bknix/build/dmaster/web/sites/all/modules/civicrm');
$routes = [
  [
    'path' => 'civicrm/contact/view/activity',
    'page_callback' => ['CRM_Activity_Page_Tab', 'run'],
    '_source' => include(CORE . '/CRM/Core/xml/Menu/Contact.xml'),
    // We don't really care to include(), but PhpStorm supports drill-down on static include() calls, so this facilitates code-navigation.
  ]
];
```

Issue #314, opened 2023-11-15
Support setting-admin mixin


mattwire:
It works fine when I copy manually from civicrm core... but civix doesn't know about setting-admin mixin yet.

PR #315, opened 2023-11-16
Add setting-admin@1. Use some extra patches to achieve 5.27+ compatibility

totten:
This adds support for `setting-admin@1` (5.68's https://github.com/civicrm/civicrm-core/pull/27569). 

To enable it, one runs:

```
civix mixin --enable=setting-admin@1
```

This creates the following behavior:

* Read `*.setting.php` for settings in this extension
* Create a permission `administer {MYEXT}`
* Create a route `civicrm/admin/{MYEXT}`
* Create a nav-item for `civicrm/admin/{MYEXT}`

## Compatibility

I did some evaluation about version-compatibility. The original implementation of `setting-admin@1` has some  issues -- e.g. `_ts()` requires v5.64, and the chosen namespace would be problematic with the polyfill-loader (before v5.45). So I made a few updates to expand compatibility, generated a couple extensions (*with this mixin*), and spot-checked the admin-page on several versions (5.27, 5.39, 5.45, 5.68).

Note: There is a related update https://github.com/civicrm/civicrm-core/pull/28170. (The relative timing is not super-sensitive. But we should merge both.)

--

See also #314 (ping @mattwire).

### DIFF
--- a/mixin-backports.php
+++ b/mixin-backports.php
@@ -69,6 +69,14 @@ return [
     'provided-by' => '5.51.beta2',
     'minimum' => '5.51', /* No point in deploying no systems that lack civicrm-core:693067e365915ce280217047009c9e87d70d0719 */
   ],
+  'setting-admin@1' => [
+    'version' => '1.0.1',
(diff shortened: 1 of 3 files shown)


totten:
civibot, test this please

totten:
civibot, test this please

PR #316, opened 2023-11-23
Readme - Remove clutter

colemanw:
Removes a bunch of boilerplate that clutters up every extension's readme file.

See https://chat.civicrm.org/civicrm/pl/5z13tkz8o7bypxrp8cis4xec7e

I think the aspiration of this boilerplate is good, but in practice:

- System requirements change over time (and are redundant with the info.xml file).
- Screenshots quickly become outdated (if anyone bothers to replace the placeholder image at all).
- Installation instructions can be centralized and don't need to be copied into every extension's readme file.


colemanw:
@artfulrobot thanks, added.

aydun:
That's all good except that the docs page doesn't include the detailed information for `cv dl` or `git clone`   Would be helpful not to loose that.

colemanw:
@totten let's merge this!

totten:
civibot, test this please

totten:
civibot, test this please

totten:
There was an early test fail that was relevant, but I think that the latest one is just HTTPD flakiness.

PR #318, opened 2023-12-03
Add helpers for naming classes like "Civi\MyExt\Foo"

totten:
This is a step toward implementing `civix generate:service`.


totten:
civibot, test this please

In the [first run](https://test.civicrm.org/job/CiviCRM-Civix-PR/58/), there were some reported failures. These look unrelated (*we have some on-going flakiness when E2E tests hit HTTPD*). We'll see how the next run does.

PR #319, opened 2023-12-03
Add command `generate:service`

totten:
Add a command to generate ["services" in the "service container"](https://docs.civicrm.org/dev/en/latest/framework/services/). The code-style is based on [automatic registration](https://docs.civicrm.org/dev/en/latest/framework/services/#automatic) in v5.55+. By default, it implements `EventSubscriberInterface` (since this is common).

## Usage

```bash
## Generate service (without specifying a name)
civix generate:service

## Generate service (with specific name)
civix generate:service myextension.foo
```

Note: If you run `civix` interactively, it will prompt to confirm/revise the name of the service and PHP class.

## Output

On a demo extension (named `civix_addsvc`), here's the console output:

Screenshot 2023-12-03 at 6 47 17 PM

Which generates a file like:

```php
<?php
use CRM_CivixAddsvc_ExtensionUtil as E;
use Civi\Core\Service\AutoService;
use Symfony\Component\EventDispatcher\EventSubscriberInterface;

/**
 * @service civix_addsvc.myService
 */
class CRM_CivixAddsvc_MyService extends AutoService implements EventSubscriberInterface {

  public static function getSubscribedEvents(): array {
    return [
      // '&hook_civicrm_alterContent' => ['onAlterContent', 0],
      // '&hook_civicrm_postCommit::Contribution' => ['onContribute', 0],
      // TIP: For hooks based on GenericHookEvent, the "&" will expand arguments.
    ];
  }

  // /**
  //  * @see \CRM_Utils_Hook::alterContent
  //  */
  // public function onAlterContent(&$content, $context, $tplName, &$object) { ... }

  // /**
  //  * @see \CRM_Utils_Hook::postCommit
  //  */
  // public function onContribute($op, $objectName, $objectId, $objectRef = NULL) { ... }

}
```

## Comment

This generator supports the `Civi\` namespace, but (for existing extensions) the `CRM_` namespace is the default. Note that:

* The default namespace comes from `info.xml`
    ```xml
    <civix>
       <namespace>CRM/Myextension</namespace>
    </civix>
    ```
* The command prompts interactive users, giving them the chance to rename the class.
* For non-interactive usage, the command accepts option `--naming=Civi` or `--naming=CRM`.


totten:
Test failure unrelated.

PR #320, opened 2023-12-05
Export - warn but don't block exporting duplicate managed entity

colemanw:
Per discussion, sometimes there's a legit reason to export a managed entity that already exists.

The use-case here is that Afform autogenerates a managed entity for navigation items that belong to the form, but civix exports them as managed entities in their own right. This isn't a real conflict but civix thinks it is.


totten:
civibot, test this please

totten:
Test failure unrelated

PR #322, opened 2023-12-09
Export - exclude readonly fields from .aff.php file

colemanw:
Noticed in some of the sprint PRs a `modified_date` was sneaking into the `.aff.php` file.

Excluding all `readonly` fields gets rid of that, and also  don't need to hard-code unsetting "search_displays" because it's also `readonly` in the metadata.

### DIFF
--- a/src/CRM/CivixBundle/Command/AddManagedEntityCommand.php
+++ b/src/CRM/CivixBundle/Command/AddManagedEntityCommand.php
@@ -168,10 +168,12 @@ with most existing extensions+generators.
 
     // Export meta file
     $meta = $afform;
-    unset($meta['name'], $meta['layout'], $meta['search_displays'], $meta['navigation']);
-    // Simplify meta file by removing values that match the defaults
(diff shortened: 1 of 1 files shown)


colemanw:
@totten not sure how to trigger CI on this repo.
retest this please

colemanw:
@totten [CI errors](https://test.civicrm.org/job/CiviCRM-Civix-PR/74/BKPROF=min,BLDTYPE=drupal-clean,CIVIVER=master,SUITE=make-snapshots,label=bknix-tmp/) due to php version:

> Fatal error: Composer detected issues in your platform: Your Composer dependencies require a PHP version ">= 7.4.0". You are running 7.3.33.

totten:
@colemanw Yup. Future test runs here should use php74.

This change sounds reasonable 👍.

Note that I was doing some reorganizing while offline, which led a fairly big patch-set - which would have a merge-conflict on this file. (*It moves this bit of code, although it doesn't substantively change those lines.*) I've copied/ported this change to the other branch (#325) -- seems like the easiest way to avoid MC.

PR #325, opened 2023-12-11
"export" - Convert to helper-style. Refine afform exports. Prettier console messages.

totten:
## Gist (Non-technical)

Reorganize the `export` command. Fix a bug with re-exporting existing files.

## Gist (Technical)

Move helper functions from `AddManagedEntityCommand` to `Upgrader`

## Discussion

This is a stylistic change and a functional fix.

* The stylistic issue is that civix currently has two code-styles in the code-generation logic.
    * (*older*) Most of the `generate:xxx` commands work with `CRM\CivixBundle\Builder\*` classes (with an OOP hierarchy). 
    * (*newer*) Most of the auto-upgrade steps (`upgrade/*.up.php`) use `CRM\CivixBundle\Upgrade::*`  functions.
    * When working on the upgrade-steps, I found it a bit tedious to use the `Builder` contract for everything.
    * The basic advantages of the newer style are: (1) it has access to more utilities (`$this->infoXml`, `$this->io`, and so on*), (2) it's easier to incorporate prompts/user-feedback, (3) it's easier to compose/mix multiple steps,
    * Going forward, I'd like to switch the general UX style of civix to be more interactive-by-default. We get some miscommunication because important options (like `<compatibility>`) have been hidden-by-default. The "Upgrader" is generally more agreeable with this style of interaction. (*It's not perfect either. Maybe it evolves more. But it's comparatively easier...*)
    * There is small functional compromise here -- the older style allows you plan a batch of changes and then execute them together. So theoretically you can bailout before executing the batch. However, I don't think this is particularly valuable in practice. (*This falls well short of "atomicity"...  And the protection provided by `git` is more effective...*) 
* The functional issue is that `civix export` didn't seem to write updates correctly. (*It said that it was writing new files, but the content never changed.*) I strongly suspect this had to do with how the `Builder` lifecycles were arranged (`loadInit()`, `save()`, etc) . But it didn't seem worthwhile to dig into if the goal is generally change the style anyway.

## And also...

* Includes a tweak for the afform exporter to omit unnecessary properties (side-port of @colemanw's #322) 
* Includes a cleanup of the console messages to  show prettier paths (aka #324).

### DIFF
--- a/src/CRM/CivixBundle/Builder/Content.php
+++ b/src/CRM/CivixBundle/Builder/Content.php
@@ -2,6 +2,7 @@
 namespace CRM\CivixBundle\Builder;
 
 use CRM\CivixBundle\Builder;
+use CRM\CivixBundle\Utils\Files;
 use Symfony\Component\Console\Output\OutputInterface;
 
 /**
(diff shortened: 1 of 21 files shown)


colemanw:
Looks good, thanks @totten

PR #329, opened 2024-02-08
Set table collation to match core tables

colemanw:
Overview
----------
Fixes inconsistent collation between extension tables (including core extensions like SearchKit) and core tables.
See discussion at https://chat.civicrm.org/civicrm/pl/g3jt55akb3rk8dzgtqcr3bjjqy

Before
-------
- All core tables are `ENGINE=InnoDB DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci ROW_FORMAT=DYNAMIC;`
- All extension tables are `ENGINE=InnoDB`

After
-------
- All core **and** extension tables are `ENGINE=InnoDB DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci ROW_FORMAT=DYNAMIC;`

### DIFF
--- a/src/CRM/CivixBundle/Command/AddEntityBoilerplateCommand.php
+++ b/src/CRM/CivixBundle/Command/AddEntityBoilerplateCommand.php
@@ -218,12 +218,12 @@ class AddEntityBoilerplateCommand extends AbstractCommand {
     return [
       'name' => '',
       'attributes' => '',
-      'tableAttributes_modern' => 'ENGINE=InnoDB',
+      // These are the attributes used by the vast majority of tables in CiviCRM core (except maybe log or archive tables).
+      // FIXME: It's not great that this is hard-coded here, but the alternative of leaving this out would be worse
+      // as the default collation setting in MySQL could be anything & we'd rather be consistent with Civi core tables.
+      'tableAttributes_modern' => 'ENGINE=InnoDB DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci ROW_FORMAT=DYNAMIC',
+      // Probably unused
       'tableAttributes_simple' => 'ENGINE=InnoDB',
-      // ^^ Set very limited defaults.
-      // Existing deployments may be inconsistent with respect to charsets and collations, and
-      // it's hard to attune with static code. This represents a compromise (until we can
-      // rework the process in a way that clearly addresses the inconsistencies among deployments).
       'comment' => '',
     ];
   }


colemanw:
@totten it seems that this humble patch has sparked a [much bigger discussion](https://lab.civicrm.org/dev/core/-/issues/4999) full of utopian dreams and aspirations...
but meanwhile this is still causing problems in the real world. For example, [this latest report on the symptoms](https://github.com/systopia/de.systopia.eck/issues/115).

So while I really do like the direction of the utopian discussions, how about we merge this in the meantime.

totten:
@colemanw OK, I want this to makes sense. On general principle, the thing to consider is whether this will be bouncy (*fix one environment, break another*). My best rationalization in favor it is:

* The current rules for character-set/collation in `civicrm-core` originated in 5.33 (https://github.com/civicrm/civicrm-core/commit/828c69155ec55d6abf50db8a8fc0f92e838faacf#diff-f73a612034e9487c9dcca2569aeb7f42ce8e03a8932fbf51d4a5ad90e14c388eL5).
* Since `civix` moved to `<upgrader>` tag, the generated entities have required `civicrm-core`@`5.38`+.
* So if you're generating ext-SQL, then you're running on 5.38+, and the collations match... right?

**But** looking at the change from 5.33... and looking at current upgrader/status-checks... it doesn't look like we ever fixed the character-sets/collations of databases on historical data. If you originally installed CiviCRM 4.6 or 5.0 or 5.10 (*or anything  up to 5.33*), then all your tables were initialized with `utf8`/`utf8_unicode_ci`, and nothing ever prompted you change. So if you install an extension that forces `utf8mb4`/`utf8mb4_unicode_ci`, then the schema will be mismatched. Right?

Doesn't this just shift the _bad-schema-situation_ to different users?

(cc @seamuslee001 b/c I think you all followed the history of utf8(mb4) more)

-------

Aside: I think I have a better way to maintain/hook-into the `CRM_*_Upgrader` and provide auto-generation. Pushing on that more today...

colemanw:
> Doesn't this just shift the bad-schema-situation to different users?

Well, I think that ship might have already sailed. Because if we wanted to avoid forcing the collation when we add new tables, then why did we do this?

https://github.com/civicrm/civicrm-core/blob/b11418429c627738ffaa9f2face1d6c68af3bd61/CRM/Upgrade/Incremental/sql/5.47.alpha1.mysql.tpl#L11

And this?

https://github.com/civicrm/civicrm-core/blob/3a3d7c89e1f4d9d45fb21a6a9bf769c4d4a361da/CRM/Upgrade/Incremental/sql/5.50.alpha1.mysql.tpl#L20

mattwire:
@colemanw @totten I remember a time when we had specified UTF8 in lots of extensions and it caused major problems when switching systems to utf8mb4 (because the SQL forced the extension to UTF8 instead of using the database default which was set to utf8mb4). So the decision at the time was to specifically *exclude* the database collation from the SQL and rely on you setting the default collation on the database once only. That allows for an easy future change of collation and also makes it easier if someone decides they need to run with another collation. I think there may have been some issues with forcing ROW_FORMAT as well. @artfulrobot I seem to remember you having some thoughts around this.

mattwire:
I'm all in favour of 2 and 3. Just not sure that 1. is a good idea. I have over 100 extensions on my local dev environment and *none* of them specify database charset/collation in their SQL except some really old ones that specify utf8. I'd rather not have to go through a round of updating sql files only to be replaced with 2 and 3 and potentially require another round of updates.

mattwire:
An aside to this, and probably more important as it's certainly tripped me up a few times is when extensions have tables that don't start with `civicrm_`  -eg. `cividiscount_` and `civirule_` as these tend to get ignored by things like `System.utf8conversion` and detailed logging.

colemanw:
Cool.

I guess I didn't see this PR so much as "require a bunch of extra work from ext maintainers" as "prevent extra work or regressions" because at this point we've [patched all core extensions](https://github.com/civicrm/civicrm-core/pull/29384) to be consistent with core but if we don't merge this update and someone reruns `civix generate:entity-boilerplate` then those fixes will be automatically undone.

I think it's otherwise fine for you to sit on your 100 extensions and not update them until [the full fix](https://github.com/totten/extlib) is ready.

colemanw:
> when extensions have tables that don't start with civicrm_ -eg. cividiscount_ and civirule_ as these tend to get ignored

Gawd why do we bother maintaining a [complete list of all dao tables](https://github.com/civicrm/civicrm-core/blob/4663912c6820c5f5f98d5b5aeeed06b005cdbad4/CRM/Core/DAO/AllCoreTables.php#L18) when stuff like that just ignores it in favor of string-fu?

colemanw:
> @artfulrobot I save megagigapetaexa bytes

Wow that's a lot! By my calculations that's `10^6 * 10^9 * 10^15 * 10^18`! 🤓 

But yea we should take into account that some super-smart admins might want to change row format and not be too restrictive about that. Or if we were super-smart maybe we could come up with some better defaults for that setting per-table.

mattwire:
> > when extensions have tables that don't start with civicrm_ -eg. cividiscount_ and civirule_ as these tend to get ignored
> 
> Gawd why do we bother maintaining a [complete list of all dao tables](https://github.com/civicrm/civicrm-core/blob/4663912c6820c5f5f98d5b5aeeed06b005cdbad4/CRM/Core/DAO/AllCoreTables.php#L18) when stuff like that just ignores it in favor of string-fu?

Yeah.. then we have things like `CRM_Core_DAO::getTableNames()` that just returns every table that starts with `civicrm_` and I think that `AllCoreTables` only returns tables linked to an entity - so possibly including non standard names?

totten:
Previously replaced by #331

PR #331, opened 2024-02-22
(dev/core#4999) Support Entity Framework v2 for extensions

totten:
Overview (*Final*)
------------

Implement support for Entity Framework v2 (EFv2; aka [dev/core#4999](https://lab.civicrm.org/dev/core/-/issues/4999)). This is a general overhaul of everything related to `generate:entity`. There's a more complete comparison of changes at: https://github.com/totten/civix/wiki/Entity-Templates

The PR was originally opened to address one major sub-problem (*installation and SQL generation*). I'm preserving that description so that the discussion remains intelligible. However, the ultimate scope of the PR evolved to cover more aspects of dev/core#4999 / EFv2. The wiki page gives a more comprehensive summation.

---

(*___Subject__: Auto-generate SQL during installation via `civimix-schema@5.72`_*)

Overview (*Original*)
-------------

For extensions that use `generate:entity` / `generate:entity-boilerplate`, this changes the SQL lifecycle:

* Remove static SQL files (`auto_install.sql` and `auto_uninstall.sql`).
* Generate dynamic SQL during installation. 
* Attune the dynamic SQL to match the existing collation of the end-user's CiviCRM schema.

ping @colemanw @mattwire @artfulrobot @christianwach

Before
---------

* `civix generate:entity-boilerplate` reads `xml/schema/**.xml` and generates SQL files (`sql/auto_install.sql` and `sql/auto_uninstall.sql`). 
    * These are static SQL files. It is not possible for the SQL files to be equally correct on all deployments, because existing deployments may have different character-set/collation flags (esp `utf8` vs `utf8mb4`).

After
-----

* During installation, we call a new helper class `CiviMix\Schema\Automatic\MyModule`
    * The helper generates SQL dynamically. It matches local character-set/collation options using the same technique as `CustomGroup`s.
    * The class is similar to a mixin. It is design to be included with `civicrm-core` and backported via `civix`. As with mixins, it will check the `<compatibility>` when deciding whether a backport is needed. At runtime, it will look to core-code and extension-code and choose the newest version.
    * The implementation can be seen [here](https://github.com/civicrm/civicrm-core/pull/29422/files#diff-653e5bcc5836a58cf977af53cd3bc31c840eb7d8467d68dda2409cc0aa0f3038R5).
* `civix generate:entity-boilerplate` does not generate SQL files.
* `civix upgrade` will prompt you to apply the following changes

    Screenshot 2024-09-17 at 4 08 43 PM

    [~~Old screenshot~~](https://github.com/totten/civix/assets/1336047/23adf5c3-7fd0-44c0-a275-cfbcf6dfc0de)

Related PRs
----------------

* This is an alternative to #329.
* The helper classes are canonically defined by https://github.com/civicrm/civicrm-core/pull/29422/.
    * This represents a *soft dependency*. The PR's can be merged in any order (or with delayed timing). But in the long run, one doesn't make sense without the other.
* The outcome of running this upgrade is demonstrated on `mosaico.git` with https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/pull/635

TODO
--------

* For new extensions, the command `generate:upgrader` should set `<upgrader>CiviMix\Schema\Automatic\MyModule</upgrader>`.
* Need to test the generated code in more environments/versions.
* Need to test as `civix.phar`. (I've been testing local git copy.)
* The test suites will probably be unhappy and will probably merit some revised coverage.


mattwire:
Nice! @totten don't let this be a blocker but just wondering if the upgrade scenario has been considered? Where a new entity is added to already installed extension and needs to install that entity on upgrade. Maybe there is a function we can call from the upgrader?

colemanw:
... or to keep it very explicit: `<autoAdd>1002</add>`

totten:
I've pushed a set of updates (but haven't updated description yet). Basically:

* Add a new helper to big E, `E::schema()`. It has the `install()` and `uninstall()` helpers. It's a good a place to add more utilities.
* Rename `CiviMix\Schema\Automatic\MyModule` to `CiviMix\Schema\MyModule\AutomaticUpgrader`
* In the future, we should look to update `AutomaticUpgrader` to include support for `<add>` or `<autoAdd>` metadata. For now, we impose some limits on `AutomaticUpgrader` to ensure that we will be able to slot-in extra revision-steps. (*Basically, as long as downstreams strictly use the `upgrade_NN()` notation, we should be able to inject some more `NN`s. But if the structure is customized... which is extremely rare... then it would be hard.*)
* Re-word the notice. Basically, the `E::schema()` helper is always going to be there (so `civimix-schema@5` will always be required). The only decision is whether you want to enable `AutomaticUpgrader`.
* Updated the related PR's (for `civicrm-core` and `mosaico`) accordingly.
* Take a pass at fixing some of the unit-tests. (There are probably still more issues, but I think got a few.)

totten:
civibot, test this please

totten:
So I rebased and switched this from the unreleased `civimix-schema@5.72` (*XML-flavor schema helper*) to the merged `civimix-schema@5.74` (*PHP-flavor schema helper*).  Quick take:

* Everything that was in the scope of the PR is... still doing what it was supposed to (i.e. it uses metadata to generate the install/uninstall SQL at runtime). 🟢 
* But goal posts move on other aspects - e.g. one needs to also address XML=>PHP conversion and `entity-type-php@1 => entity-type-php@2`. ⚠️ 
* To demonstrate on `mosaico`, I did the automatic upgrade and then manually updated those other bits. ([Branch](https://github.com/veda-consulting-company/uk.co.vedaconsulting.mosaico/compare/2.x...totten:uk.co.vedaconsulting.mosaico:2.x-autosql-php?expand=1)) And it does create new tables. 🎉

src/CRM/CivixBundle/Resources/views/Code/module.civix.php.php
```diff
@@ -82,10 +82,39 @@ public static function findClass($suffix) {
...
+<?php if (\Civix::checker()->hasMixinLibrary()) { ?>
+pathload()->addSearchDir(__DIR__ . '/mixin/lib');
+<?php } ?>
+<?php if (\Civix::checker()->hasMixinLibrary('civimix-schema@5')) { ?>
+spl_autoload_register('_<?php echo $mainFile ?>_civix_class_loader', TRUE, TRUE);
+
+function _<?php echo $mainFile ?>_civix_class_loader($class) {
+  // This allows us to tap-in to the installation process (without incurring real file-reads on typical requests).
+<?php $_clPrefix = 'CiviMix\\Schema\\' . \CRM\CivixBundle\Utils\Naming::createCamelName($mainFile) . '\\' ?>
+  if (substr($class, 0, <?php echo strlen($_clPrefix) ?>) === <?php var_export($_clPrefix) ?>) {
```
colemanw, line 109: This is really brittle. How about:
```suggestion
  if (str_starts_with($class, <?php var_export($_clPrefix) ?>)) {
```

totten: "Really brittle" is subjective:

* Yeah, if we're talking about code that a developer edits by hand, then `substr(str,0,n)`  is more brittle than `str_starts_with()`. (*I definitely prefer using `str_starts_with()`...*)
* But `*.civix.php` is machine-generated glue, so the human-editing isn't a big factor. As generated glue code, its interoperability is a consideration. ("So Bob tried the extension on php7+Civi 5.53; he says it works. Carol on php8+Civi 5.52 also says it works. Then Dave tried it on same versions and it crashed -- because his specific permutation, php7+Civi 5.52, lacks the polyfill...") From compatibility POV, `str_starts_with()` is more brittle than `substr()`.

So I'd ordinarily agree, but in this context it's easier to use the Old Ways.

colemanw: @totten somehow I knew you would say that... so how about a compromise:
```suggestion
  if (strpos($class, <?php var_export($_clPrefix) ?>) === 0) {
```

totten: @colemanw Yup, that sounds good.

colemanw:
@totten what's the status of this?

colemanw:
@totten I've just finished converting the rest of our core extensions to use this.

src/CRM/CivixBundle/Resources/views/Code/entity-dao.php.php
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * DAOs provide an OOP-style facade for reading and writing database records.
+ *
+ * DAOs are a primary source for metadata in several versions of CiviCRM (<5.75)
+ * and are required for some subsystems (such as APIv3).
+ *
+ * This stub provides compatibility. It is not intended to be modified in a
+ * substantive way. However, you may add comments and annotations.
+ */
+class <?php echo $daoClassName ?> extends CRM_Core_DAO {
```
colemanw, line 16: Unfortunately this file probably won't do anything but cause errors. Since it doesn't extend `CRM_Core_DAO_Base` it's kinda like implementing an interface without all the required methods (CRM_Core_DAO doesn't have a *real* interface, but there are certain functions that are expected to exist).

totten: Oh, right. We'd talked about the need for a backport of `CRM_Core_DAO_Base`. Haven't gotten that far yet...

totten: (Don't want to think about that backport yet -- should probably move toward testing `master`/ `CRM_Core_DAO_Base` in the interim.)

totten: OK, as of a5ecaad7cde99c59 (using `CRM_Core_BAO_Base`), I think a bunch of random issues should be addressed. Tests pass for all suites involving `5.74` and `master`. Then, as you'd expect, it fails for everything older - because we don't have the substitute for `CRM_Core_BAO_Base` yet.

So that's the next on the queue... and then circle-back to manually test some real-world upgrades... and then (*knock on wood*) maybe that's it...

totten: @colemanw So this is a version of `CRM_Core_DAO_Base` that seems run on older versions (https://github.com/civicrm/civicrm-core/commit/1bcfda4044b46119a8ac92fc841cf35da6ada82f) -- at least, I did light testing with core@5.51, `cv en MYEXT`,  `cv api4 MyEntity.get`, and `cv api4 MyEntity.create`.

I've been going back-forth about how to wire that implementation into `civix` templates -- or, more specifically, whether it's better to:

1. Lean on "reloadable" conventions for DAO-base (*i.e. always using the dynamic class*). 
    * https://github.com/totten/civix/commit/b122378469b8fe20c30e8594ffca80fdcea3701b
2. Lean more on "static" conventions for DAO-base (*i.e. for users on 5.74+, the aliases will be static and visible to the IDE; only use the magic-base for <5.74*). 
     * https://github.com/totten/civix/commit/352a2b48d589bc817a284c42bd94d8ac4829b867

But I would like to get some test-results, so I'm gonna go ahead and push one of them onto this branch.

totten:
Ah, blerg. There's a test-interaction problem (i.e. tests pass individually but fail in bulk). E2E tests look like this:

```php
   $this->civixGenerateModule(static::getKey(), [
      '--compatibility' => '5.69',
    ]);
    $this->civixGenerateUpgrader();
    $this->civixGenerateEntity('Bread');
    $this->civixGenerateEntity('Sandwich');
 ```

* These steps actually run in the same PHP process. This is good for debugging/inspection -- but bad for fidelity. In real usage, they run as separate PHP processes.
* If you run several tests in the same process, then the relevant files/functions only load once. (*The first variant of `function xyz() {...}` stays in memory.*)
* This hasn't been a problem before because...
    * The code-generators were traditionally/typically additive (*each test/step makes a new extension/file/function/class*).
    * The code-generators were traditionally/typically passive (*they don't necessarily boot CiviCRM; they just fiddle with strings*).
* But now...
    * It's starting to lean more on conditional construction (*and regeneration*) of `*.civix.php`. Steps aren't just additive -- they can change the content of `*.civix.php`. (*This PR is a good example.*)
    * It's starting to incorporate more active functionality from CiviCRM -- so some steps require booting (*while other steps don't*).
* The overall effect is chaotic/painful to trace.
* What's probably needed is to change the test-structure to improve isolation. We have a couple high-leverage points:
    * phpunit's `--process-isolation` puts each test-case in its own process. But it's not quite realistic -- individual steps would still share the same process.
    * `CivixProjectTestTrait::civix()` calls Symfony's `CommandTester`. Swap that with something similar that uses sub-processes. This could reveal some *other* hidden interactions, but it's probably better in the long run.

tldr: I need a separate PR to change the testing process.

src/CRM/CivixBundle/Resources/views/Code/entity-dao.php.php
totten: OK, it's using the more static-looking convention (`CRM_Myext_DAO_Base` presents as the base-class; internally, its mapped to either `CRM_Core_DAO_Base` or `CiviMix\Schema\MyExt\DAO`.) It took a few more revisions on other topics, but that arrangement does seem to be passing.

colemanw:
@totten this is looking good!

totten:
Fixed a "WIP" commit that was outstanding. We had a clean test-run. I also did some local trials like this:

* Install 5.45. Make extension `foo545`. Add entity `Foo545`. Install this anew on 5.45 (and also 5.69 and 5.78).
* Install 5.69. Make extension `foo569`. Add entity `Foo569`. Install this anew on 5.69 (and also 5.78).
* Install 5.78. Make extension `foo578`. Add entity `Foo569`. Install this anew on 5.78.

I also went back and re-tried `civix upgrade` and `civix convert-entity` on `mosaico.git` (with core@5.69). Ran into a couple issues. In my basic testing, the latest couple commits deal with these. `civix upgrade` will prod you over to EFv2.

Hopefully (*knocks on wood*) test-suites will still be happy...

totten:
FIXME: Upgrade/convert step also needs to generate skinny DAOs.

totten:
The upgrade/convert step will now re-generate DAOs -- and delete the old XML schema files.

Unless the test-suite has some complaints with this revision, then I think the PR may be ready to merge....

colemanw:
Looks good @totten 

As you probably remember, my vision is for these thin DAOs to be temporary, and as the core EntityRepository matures, the API will switch to [calling `$entity->writeRecord()` directly](https://github.com/civicrm/civicrm-core/pull/30995) and the DAO itself will be obsolete. But I think this is good for now.

colemanw (approved):

PR #337, opened 2024-03-12
Deprecate api command

colemanw:
APIv3 is officially deprecated so civix should reflect that.

Also adds `@deprecated` annotation to all deprecated commands.

### DIFF
--- a/src/CRM/CivixBundle/Command/AddAngularPageCommand.php
+++ b/src/CRM/CivixBundle/Command/AddAngularPageCommand.php
@@ -12,6 +12,9 @@ use CRM\CivixBundle\Builder\Info;
 use CRM\CivixBundle\Builder\Template;
 use CRM\CivixBundle\Utils\Path;
 
+/**
+ * @deprecated
+ */
 class AddAngularPageCommand extends AbstractCommand {
 
   protected function configure() {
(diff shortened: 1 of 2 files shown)


totten:
Wouldn't it make sense to switch this to the APIv4 equivalent?

One thing - APIv4 is in a funnier position because of how `Civi\Api4\MyEntity.php` and `Civi\Api4\Action\MyEntity\MyAction.php` relate. IIRC:

* When you generate an `action` class, you *should* update the `entity` class to declare a stub.
* But only if your extension owns the entity. You skip the stub if entity is owned by someone else.
* Adding the stub function requires modifying the PHP source. That can be fiddly, but there's a big mitigating factor -- you don't have to worry about pre-existing functions. I'd wager 98% cases would work with a little regex to inject the new function at the end of the file (*right before the terminal `}`*).

colemanw:
@totten are you saying the `generate:api` command should be changed to create Api4?

> When you generate an action class, you *should* update the entity class to declare a stub.

Should is a strong word. You don't have to. But at least adding a `@method` annotation is nice. So I guess if we're going to the trouble of that we might as well add the function. Injecting it at the end of the file ought to work fine.

PR #338, opened 2024-03-12
AddEntityCommand - Stop generating APIv3

colemanw:
It's been 5 years since this command was switched to v4 as the default.
Api3 is officially deprecated in core.
It's time.


totten:
In terms of end-result for UX, this seems OK. 

Test data is expired, so I can't see what failed. And it needs a rebase. My main question would be about the implications for upgrade-testing. But if it passes, then it passes.

colemanw:
@totten I assume the tests are expecting a v3 api to be generated?

totten:
Yeah, I'm digging back in to see. There are basically 3 kinds of tests:

* `scripts/make-snapshots.sh` is a broad/cross-cutting thing which calls a bunch of generators... and asserts that the extension is generally runnable
    * *For this, I think we can just stop calling api3 variants.*
* `tests/e2e/` are (for the most part) narrower tests of specific generators. They have more fine-tuned assertions
    * *I think this already switched to api4 (in keeping with the existing defaults of `generate:entity`).*
* `tests/e2e/SnapshotUpgradeTests` goes through a library of previously-generated extensions... and calls `civix upgrade`... and asserts each extension is generally runnable.
    * This is the part that needed more of a think.
    * So... if you have an existing extension which includes api3 stuff... then yes, you should still be able to run `civix upgrade` (even if we're not going to generate new api3 code). So the tests should continue running against those extensions.
    * But actually, it looks like it's time to add to the library. The library has examples of pure-api3 and hybird-api3/api4... but no examples of pure-api4.

I'll push some test updates when I have a few more snapshots to include.

PR #339, opened 2024-03-12
Docs - add note on running from source

colemanw:


totten:
@colemanw I don't know if there's a better way to present this, but ... there are a lot of ways to download. The README has this section:

* https://github.com/totten/civix/tree/master?tab=readme-ov-file#download

which links to

* https://github.com/totten/civix/blob/master/doc/download.md#src-unix

Would it be clearer to emphasize that `civix.git` means "running the source code", eg

* [Install `civix.git` (*source*) as system-wide tool (Linux/BSD/macOS)](https://github.com/totten/civix/blob/master/doc/download.md#src-unix)
* [Install `civix.git` (*source*) as system-wide tool (Windows)](https://github.com/totten/civix/blob/master/doc/download.md#src-win)
* [Install `civix.git` (*source*) as project tool (composer)](https://github.com/totten/civix/blob/master/doc/download.md#src-composer)

colemanw:
I guess I'm just trying to state what you thought was the obvious. As the author of civix you thought it was self-explanatory that civix runs from source code without any compilation. All the links you just shared seem to take that fact for granted. But I had always thought there was some sort of compile step needed to turn it into an executable shell script and for years I've been faithfully following the instructions on this page to run the `box` command if I want to test out a patch for civix. Maybe I'm the only one.

PR #340, opened 2024-03-12
Fix various warnings on PHP 8.2 / PHPUnit 9

totten:


totten:
This is passing in most permutations. The remaining failures are *probably* a consequence of incompatibilities between php@8.2 and core@5.57. It's annoying to show formally. But I think this is better in than out.

PR #342, opened 2024-03-14
Add convert-entity command

colemanw:
Adds command to convert entities from legacy xml/schema to current .entityType.php format.

Used to generate the `entityType.php` files in https://github.com/civicrm/civicrm-core/pull/29472


totten:
civibot, test this please

totten:
The PR doesn't directly change of the existing workflows. We can merge it for now and use it to help test other questions/steps in the migration path.

PR #343, opened 2024-03-21
Add entity-types-php@2

colemanw:
Adds new mixin from https://github.com/civicrm/civicrm-core/pull/29771

### DIFF
--- a/mixin-backports.php
+++ b/mixin-backports.php
@@ -50,7 +50,7 @@ return [
     'remote' => 'https://raw.githubusercontent.com/civicrm/civicrm-core/5d5c59c9453fabf1c3f709c7b8e26c680312d9a8/mixin/entity-types-php@2/mixin.php',
     'local' => 'extern/mixin/entity-types-php@2/mixin.php',
     'provided-by' => '5.73',
-    'minimum' => '5.73',
+    'minimum' => '5.51',
   ],
   'menu-xml@1' => [
     'version' => '1.0.0',
--- a/mixin-backports.php
+++ b/mixin-backports.php
@@ -44,6 +44,14 @@ return [
     'provided-by' => '5.57.beta1',
     'minimum' => '5.45',
   ],
+  'entity-types-php@2' => [
+    'version' => '2.0.0',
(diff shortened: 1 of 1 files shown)


mixin-backports.php
```diff
@@ -44,6 +44,14 @@
     'provided-by' => '5.57.beta1',
     'minimum' => '5.45',
   ],
+  'entity-types-php@2' => [
+    'version' => '2.0.0',
+    'sha256' => 'ea98aa9ec3a4ee2c0c2f19cf264f20db344d4d25e2dc714e0ba45cccb1cf6c56',
+    'remote' => 'https://raw.githubusercontent.com/civicrm/civicrm-core/5d5c59c9453fabf1c3f709c7b8e26c680312d9a8/mixin/entity-types-php@2/mixin.php',
+    'local' => 'extern/mixin/entity-types-php@2/mixin.php',
+    'provided-by' => '5.73',
+    'minimum' => '5.45',
```
totten, line 53: Should we temporarily mark this as `minimum=>5.73`?  I guess we're still shaking-out that detail.

colemanw: Fine with me! I'm still struggling to see the value of backporting this change.

colemanw:
@totten updated the minimum.

totten:
Ah, erg. [E2E\MixinMgmtTest.testEnableAllDisableAll](https://test.civicrm.org/job/CiviCRM-Civix-PR/109/BKPROF=min,BLDTYPE=drupal-clean,CIVIVER=master,SUITE=run-tests,label=bknix-tmp/testReport/junit/(root)/E2E_MixinMgmtTest/testEnableAllDisableAll/) ==> it makes sense for that assertion to fail... The specific number is a bit arbitrary, but it would have to bounce around with this.

TBH, I'm not really sure the benefit to separating this commit from the other parts of the project... (Like I view it opportunistically - if separating makes it easier, then well-and-good.)

totten:
civibot, test this please

colemanw:
@totten FIFY

totten:
Yeah, fair enough. `hook_entityTypes` is totally a valid thing in 5.51... and storing the `*.entityType.php` in `./schema` would be too.

There's other plumbing for the overall set of changes/expectations, but that needs to be sorted out in other PRs. Don't need to block this part.

PR #344, opened 2024-03-21
Fix php notices

colemanw:
Fixes test failures on `max`.

### DIFF
--- a/src/CRM/CivixBundle/Builder/Collection.php
+++ b/src/CRM/CivixBundle/Builder/Collection.php
@@ -8,6 +8,12 @@ use Symfony\Component\Console\Output\OutputInterface;
  * A collection of builders
  */
 class Collection implements Builder {
+
+  /**
+   * @var array
+   */
+  public $builders;
+
   public function __construct($builders = []) {
     $this->builders = $builders;
   }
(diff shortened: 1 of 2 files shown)


colemanw:
Well crap I started this PR because of all the php warnings causing test fails and then realized you've already opened a similar PR #340 @totten can you please merge that other one so the tests can run?

totten:
@colemanw To your main point, yes, civix does need these kinds of updates. And PR #340 is actually further along for doing them. And what I really want is for the test-matrix to show that #340 achieved its goal wrt to civix/php82. But it can't quite do that because it's wrapped up with the question of *core's compatibility*. 

Put another way - even if (*as I believe*) #340 fixes compatibility for civix/php82, it still gets flagged as red because civix+cv support multiple versions of core (including core@5.57, which doesn't support php82). For testing downstream tools like cv/civix, the test-environments need permutations like:

* core@esr-2 + php@8.1
* core@rc + php@8.2
* core@master + php@8.2

(*It's usually not that bad -- we go long periods keeping the requirements flat for many versions. But we're in a transitional period on core's requirements. There are others way to haggle -- maybe drop 5.57 (*even though it's actually still compatible*)? But 5.63 is probably no better. And the whole thing will recur. Usually, we get a good 6-18 months... but with current php83 chatter, I don't think we're looking a short window here. I think it's better to fix the root problem -- supporting better permutations of testing core x php*)

PR #345, opened 2024-04-19
Convert entity fixes

colemanw:
Fixes some stuff I noticed when looking at the generated entityType.php files.


totten:
The corresponding core update (which demonstrates the effect of the patch) looks good. 👍

PR #348, opened 2024-05-08
mixin-backports - Update smarty adapters to v1.0.2

totten:
Update to reflect the pending update in https://github.com/civicrm/civicrm-core/pull/30148

Note: The existing 'minimum' was '5.27'. I re-tested that version locally, and Smarty still worked fine.

cc @colemanw


colemanw:
@totten 👍🏻 this can be merged.

PR #349, opened 2024-05-11
Convert-entity - 4 fixes

colemanw:
~~Three~~ 4 fixes to the entity generation, used to fix the test failures in https://github.com/civicrm/civicrm-core/pull/29991

This fixes discrepancies between the old core/CodeGen and the new civix/convert-entity output.


totten:
Woot.

PR #350, opened 2024-05-15
Update cv-lib version to match cv's version

demeritcowboy:
It seems like the cv here should be the same version as cv.


totten:
Good point @demeritcowboy. And the tests pass, so it should be good.

PR #351, opened 2024-05-25
info.xml: change <maintainer> to <authors>

### LINKED ISSUE #208: Have CiviX generate `<authors>` info.xml instead of `<maintainer>`
We should look at adding support for `<authors>` to the Civix workflows as well as/instead of Maintainer.

semseysandor:
closes #208

### DIFF
--- a/src/CRM/CivixBundle/Builder/Info.php
+++ b/src/CRM/CivixBundle/Builder/Info.php
@@ -25,10 +25,12 @@ class Info extends XML {
     $xml->addChild('name', $ctx['fullName']);
     $xml->addChild('description', 'FIXME');
     // urls
-    $xml->addChild('license', isset($ctx['license']) ? $ctx['license'] : 'FIXME');
-    $maint = $xml->addChild('maintainer');
-    $maint->addChild('author', isset($ctx['author']) ? $ctx['author'] : 'FIXME');
-    $maint->addChild('email', isset($ctx['email']) ? $ctx['email'] : 'FIXME@example.com');
+    $xml->addChild('license', $ctx['license'] ?? 'FIXME');
+    $authors = $xml->addChild('authors');
+    $maint = $authors->addChild('author');
+    $maint->addChild('name', $ctx['author'] ?? 'FIXME');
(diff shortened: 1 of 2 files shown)


totten:
Thanks for resurrecting this topic @semseysandor.

Regarding some comments/data-points from #208 discussion:

* The `<authors>` tag was introduced in 5.36 (https://github.com/civicrm/civicrm-core/pull/19684). Older versions won't display authors/maintainers properly.
    * This patch only affects the defaults generated on new extensions -- thus, new extensions require at least 5.36.
    * That's fine IMHO. (5.36 is several years old at this point... and it's not big increase from the current de-facto minimum of 5.27.) We just need to be a bit more consistent in presenting the fact that 5.36 is required.
    * I've added a commit to raise the minimum.
    * Alternatively, the code to support both would be about the same complexity. (`if (compat less than 5.36) use <maintainer> else use <authors>`) Maybe we go that route if the test-suite finds a problem... or if someone has a strong feeling about 5.27-5.35... but from my POV, it's just as well to require 5.36 on new exts.
* `#208` also suggested adding `<authors>` to a real/example extension to see if metadata propagates properly through the distribution-system.
    * Well, I searched to see if anyone has published an ext with `<authors>`. The `apikeys` extension has. It looks OK to me in `civicrm.org` directory (as well as the in-app UI -- before+after download). So that's good. :+1: 
       * (Slight nitpick - `apikeys` is publishing both `<maintainer>` and `<authors>`, which is a bit redundant. OTOH, that extension still list `<compatibility>` with `4.7` and `5.0`, so maybe that's why. That's their prerogative. Regardless, this PR does the prettier thing -- it only generates `<authors>`. Which seems like a better default for new exts.)

semseysandor:
Too bad `<authors>` hasn't really been catched up, i think it's a nice feature, though most of the extensions are maintained by a single maintainer...
At least new extensions will use it now.
Maybe we can add an upgrader, so `civix upgrade` will convert `<maintainer>` to `<authors>`? Then deprecate `<maintainer>` in core?

PR #352, opened 2024-05-30
mixin-backports - Update smarty adapters to v1.0.3

totten:
Port https://github.com/civicrm/civicrm-core/pull/30281

### DIFF
--- a/mixin-backports.php
+++ b/mixin-backports.php
@@ -95,18 +95,18 @@ return [
   ],
   'smarty@1' => [
     // Newer variant. Prettier name.
-    'version' => '1.0.2',
-    'sha256' => '71915fc6c8a826611495f07d9f350e23630c0a90d47cfebe61eca291fe6e7451',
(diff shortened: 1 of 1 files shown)


totten:
Since this has a fairly low `minimum` (`5.27`), and since it's a non-trivial reorg on the mixin, I did a bit more testing than what's in the matrix. e.g.

* Make an extension on `5.27` with a Smarty-based web-page and ensure it works. (*This worked fine.* 🟢)
* Run the `civix` test-suites with 5.27 and this patch. (The status-quo is that some civix features don't work on 5.27, so it had pre-existing failures.  But the patch didn't make it any worse.  🟡)
* Stress-test `smarty-v2@1.0.3` on `5.59`. (This is the oldest version where `tools/mixin/bin/test-all` has coverage for Smarty. *This worked fine.* 🟢)
* (Incidentally, I also did a s stress-test with using `hook_install` to render some `*.tpl`s. This didn't work before. In fact, `upgrades/22.12.1.up.php` and `upgrades/23.01.0.up.php` already pointed to this. 🟡 )

So overall, 1.0.3 looks to be as good as 1.0.2.

Will merge after 30281.

totten:
Rebased to include the final revision of smarty@1.0.3. I don't think any of the late changes were version-relevant.

PR #362, opened 2024-09-13
Default to version 1.0.0 instead of 1.0

mlutfy:
I would like to recommend that people use `x.y.z` as versions numbers, instead of `x.y`. For now, it's just a recommendation. It makes it easier use composer (which requires semantic versioning).

Related discussion:

- https://lab.civicrm.org/documentation/docs/dev/-/merge_requests/1197#note_170192
- https://lab.civicrm.org/infra/extdir/-/issues/7


colemanw:
This looks good to me @totten ✅

totten:
Yup, +1 for encouraging SemVer

PR #364, opened 2024-09-18
Convert $attributes['multiple'] to boolean

colemanw:
Just makes more sense this way.


totten:
Cool, that makes sense to me. 👍

PR #366, opened 2024-09-19
Update references from extension aliases to the CRM_Core_Extension

eileenmcnaughton:
Prior to CiviCRM 5.52 we threw a mismash of these extensions. They all had to work everywhere because it was unpredictable. In 5.52 we made the API_Exception and CiviCRM_Api3_Exception aliases of CRM_Core_Exception. However, on the off chance people are creating extensions for really old civi versions with the latest civix it would still be handled on those older versions (we were always throwing CRM_Core_Exception all over the show)

### DIFF
--- a/src/CRM/CivixBundle/Resources/views/Code/api.php.php
+++ b/src/CRM/CivixBundle/Resources/views/Code/api.php.php
@@ -26,7 +26,7 @@ function _<?php echo $apiFunction ?>_spec(&$spec) {
  *
  * @see civicrm_api3_create_success
  *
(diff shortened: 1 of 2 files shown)


totten:
Yeah, I think this works. 👍 

The templates affected here are like demonstrative code (*some soft `@throws` and some illustrations of "how you might use an exception"*) rather than "central glue affecting all handling of exception".  Even if there are some existing extensions/historical versions which treat `API_Exception` and `CRM_Core_Exception` differently... it doesn't really affect the quality of the demonstration.

PR #367, opened 2024-09-25
composer: update totten/license-data

semseysandor:
`totten/license-data` got a new release (0.3) recently, update dependency here


totten:
Thanks @semseysandor

PR #368, opened 2024-09-25
Paths - On Windows, normalize to forward-slash

totten:
As reported by loren-bm on ([Mattermost](https://chat.civicrm.org/civicrm/pl/ruo9km1w4idz7mh34jczhpeirr)), there's a recent regression when running `civix upgrade` on Windows:

```
Executing upgrade script phar://E:/d10/vendor/civicrm/cli-tools/extern/civix.phar/upgrades\24.09.0.up.php

In Checker.php line 55:

  [RuntimeException]
  Unrecognized library: civimix-schema@5
```

The underlying problem involves the switch to `Webmozart\Glob` -- @demeritcowboy found that it expects to have paths normalized using `/`. This is not entirely unusual. Symfony Filesystem v5.4+ has a [similar normalization policy](https://github.com/symfony/filesystem/blob/7.0/Path.php#L114-L117).

This revision *should* fix the error. The approach is a little more aggressive, in that it flip-flips on the approach to normalizing paths:

* __Before__: `Path` helper tries to use native `DIRECTORY_SEPARATOR`. But it doesn't try very hard. So in practice, you probably a get a mix of delimiters.
* __After__: `Path` helper tries to consistently use `/` on both Windows and Unix. 

I've posted an example build of `civix.phar` at:

https://ephemera.civicrm.org/civix/civix.phar-2024-09-25-windows

### DIFF
--- a/src/CRM/CivixBundle/Utils/NamingTest.php
+++ b/src/CRM/CivixBundle/Utils/NamingTest.php
@@ -2,6 +2,9 @@
 
 namespace CRM\CivixBundle\Utils;
 
+/**
+ * @group unit
+ */
 class NamingTest extends \PHPUnit\Framework\TestCase {
 
   public function getFullNameExamples() {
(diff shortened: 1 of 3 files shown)


demeritcowboy:
Thanks. This works for me. 👍

PR #376, opened 2024-12-07
Handle dynamic versions

### LINKED ISSUE #375: Doesn't recognize [civicrm.majorVersion]
Trying to use the `civix export` command with a core extension I get the following error:

```sh
Enable mixin mgd-php@1.0
! [NOTE] The requirements should be increased because "mgd-php@1.0.0" requires CiviCRM 5.27.
! [NOTE] The requirements should be increased because "scan-classes@1.0.0" requires CiviCRM 5.51.
Increase minimum CiviCRM requirement from [civicrm.majorVersion] to 5.51
```

colemanw:
Fix #375

### DIFF
--- a/src/CRM/CivixBundle/Builder/Info.php
+++ b/src/CRM/CivixBundle/Builder/Info.php
@@ -15,8 +15,7 @@ class Info extends XML {
 
   public function init(&$ctx) {
     $ctx += [
-      // FIXME: Auto-detect current installed civi version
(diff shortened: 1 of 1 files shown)


totten:
Cool. Seems to work. Tried a little bit to provoke new trouble, but it just seemed to work... :)

colemanw:
Tim the troublemaker 😈😈

PR #380, opened 2025-01-15
upgrades/24.09.01: only add step to update Upgrader if there is an Upgrader

semseysandor:
If an extension has no Upgrader defined in `info.xml`, this upgrade step will tell that:
```
Update info.xml to use CiviMix\Schema\{MyExtension}\AutomaticUpgrader
```
but it won't do it.

It's just confusing for the user.
In this PR, the displayed planned actions are matched to the actual ones.


totten:
Quite right! If they don't have an `<upgrader>` or any other hints at defining entities... then we really don't need to show that message.

PR #381, opened 2025-01-16
upgrade/24.09.1, convert-entity: adds entity-types-php@2 mixin even the extension doesn't define new entities

### LINKED ISSUE #372: Unconditionally prompts upgrade to EFv2
When upgrading the [SearchKit Reports](https://lab.civicrm.org/extensions/search_kit_reports) extension from v23.02.1 => v24.09.1 I was prompted to upgrade to EFv2. That's nice, but the extension has no entity types. I typed "y", expecting nothing to happen, but `<mixin>entity-types-php@2.0.0</mixin>` was added to the `info.xml` and the shim `entity-types-php@2.0.0.mixin.php` was added to the mixin directory. I manually deleted these after the upgrade because they're irrelevant to an extension with no entities.

@totten I'm concerned that developers upgrading their extensions will wind up with a bunch of irrelevant mixins and shims in the process.

semseysandor:
`\CRM\CivixBundle\Command\ConvertEntityCommand::convertEntities()` is called regardless if there are actual legacy xml entities to convert, and it adds the `entity-types-php@2` unconditionally.

In this PR, `convertEntities` is called only if there are files to convert. Also a check is 
added to `convertEntities` to return early if it's called with no xml files.

fixes #372

### DIFF
--- a/src/CRM/CivixBundle/Command/ConvertEntityCommand.php
+++ b/src/CRM/CivixBundle/Command/ConvertEntityCommand.php
@@ -31,7 +31,7 @@ class ConvertEntityCommand extends AbstractCommand {
     Civix::boot(['output' => $output]);
     $this->assertCurrentFormat();
 
-    \Civix::io()->note("Finding entities");
+    \Civix::io()->note('Finding legacy xml entities');
 
     $isCore = $input->getOption('core-style');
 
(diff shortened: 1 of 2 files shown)


totten:
Seems good to me.

I played with running the upgrade on a few different extension scenarios, eg

* Ext with no schema/upgrader/etc; 
* Ext with custom upgrader... but no EFv1 metadata
* Ext with full EFv1 (upgrader/sql/metadata/etc)

Where it changed, it all seemed to change for the better. :+1: 

Thanks @semseysandor!

colemanw:
Great work @semseysandor !

PR #390, opened 2025-01-27
Upgrader - Fix core version downgrading & linting error

colemanw:
When running the command `civix generate:upgrader` on a core ext, I noticed two problems:
1. The compatability version was downgraded.
2. There was a linting error in the generated file.

Fixed 1 by teaching the function `raiseCompatibilityMinimum` to [work with the '[civicrm.majorVersion]' token](https://github.com/totten/civix/pull/376).
Fixed 2 by adding a newline after the `use` statement


totten:
Logic of the update seems reasonable. Tried it out a couple extensions in different  situations(core and contrib; adding an `Upgrader` and just general refresh). All seemed well.

PR #393, opened 2025-01-29
Merge `--yes` with the more conventional `--no-interaction`

### LINKED ISSUE #391: --yes ... or no?
Does the `--yes` option work?

Observe:
1. Take an extension not on EFv2
2. Run `civix upgrade --yes`
3. At several points the process will be halted for manual confirmation and you must type `y`.

totten:
Overview
-------------

Address the inconsistent behavior of `--yes` by using the (more consistent) option `--no-interaction`.

Fix #391

Before
---------

* `civix` inherits the option `--no-interaction` (`-n`) from Symfony Console.
* `civix` also defines the option `--yes`, but this is only used for 1-2 prompts. Most prompts ignore it.
    * (*I suspect it lost a lot relevance when logic was migrated from `Command` classes to the `Generator`.*)

After
------

* `civix` inherits the option `--no-interaction` (`-n`) from Symfony Console.
* `civix --yes` is a deprecated alias for `civix --no-interaction`.


colemanw:
Great, thanks @totten

PR #396, opened 2025-02-12
When will they stop being experimental?

colemanw:
Today!


totten:
I read the title, clicked the link, and waited 10 earnest seconds and then the answer was... Revealed!

Re: Test failure -- Looks unrelated. Probably a side-effect of the recent rotation of ESRs. (I'll look at that separately.)

colemanw:
Thanks!

PR #403, opened 2025-04-16
Update civixmix-schema from 5.80.2=>5.85.beta1

totten:
This should add new helpers from https://github.com/civicrm/civicrm-core/pull/32603/files.

TODO: Assess backward-compatibility. (Hopefully, the CI tests will give some clue... 🤞)


totten:
> TODO: Assess backward-compatibility. (Hopefully, the CI tests will give some clue... 🤞)

I think as currently written [this bit](https://github.com/civicrm/civicrm-core/pull/32603/files#diff-659a0c54607f257dea58c0c99a972d47f906e6050fc9bbdc4fe51524297cd12fR97) would require 5.74+, but the tests are passing because there's no coverage for `alterSchemaField()`. 

So I should probably add some coverage for the new function.

After that, I'm pretty one of these would get 5.69 (actually, 5.45) compat:

```diff
-$tableName = \Civi::entity($entityName)->getMeta('table');
+$tableName = CRM_Core_DAO_AllCoreTables::getTableForEntityName("Contact")
```

```diff
-$tableName = \Civi::entity($entityName)->getMeta('table');
+$tableName = method_exists('Civi', 'entity')
+   ? \Civi::entity($entityName)->getMeta('table')
+   : CRM_Core_DAO_AllCoreTables::getTableForEntityName("Contact")
```

But first, test-coverage...

totten:
New revision just adds some test-coverage. Should see it failing in a few places. (After that, I've got a draft revision with compatibility fix...)

totten:
Test now passes here (but with a small patch for `SchemaHelper.php`). But I'm gonna wait and merge after the canonical copy of `SchemaHelper.php` is updated.

PR #411, opened 2025-09-24
Add scan-classes mixin when adding Api4 classes

colemanw:
This used to be a soft-requirement but is now strongly encouraged.

- See https://github.com/civicrm/civicrm-core/pull/33371
- See https://github.com/totten/civix/pull/412

### DIFF
--- a/src/CRM/CivixBundle/Command/AddEntityCommand.php
+++ b/src/CRM/CivixBundle/Command/AddEntityCommand.php
@@ -69,6 +69,10 @@ explicitly.');
     $ctx['tableName'] = $input->getOption('table-name') ? $input->getOption('table-name') : Naming::createTableName($input->getArgument('<EntityName>'));
 
     $mixins = new Mixins($info, $basedir->string('mixin'), ['entity-types-php@2.0']);
+    if (in_array('4', $apiVersions)) {
+      // Needed to pick up the new Api4 class
+      $mixins->addMixin('scan-classes@1.0');
+    }
     $mixins->save($ctx, $output);
     $info->save($ctx, $output);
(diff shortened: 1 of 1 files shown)


totten:
@colemanw I think that notice should be fixed, but I think it's a red-herring. AFAICS, it only appeared once. And it's talking about the "cache warmup" which prepopulates some stuff in `composer`'s cache. (But if warmup fails... then `composer` will just download the necessary files on-demand.)

Looking at more failures, the common theme (near the end) is:

```
Enabling extension "org.example.civixsnapshot"

In MySearch.php line 7:
                                                         
  Class "CRM_Contact_Form_Search_Custom_Base" not found  
                                                         

ext:enable [-r|--refresh] [--ignore-missing] [--] [<key-or-name>...]
```


I think this reveals a latent problem:

* The `make-snapshots` job tries several scenarios for generating extensions... including the [kitchensink scenario](https://github.com/totten/civix/blob/master/tests/scenarios/kitchensink/make.sh). _That calls every generator, including `generate:search`._
* If you call `generate:search`, then you make a class which extends `CRM_Contact_Form_Search_Custom_Base` and therefore (in modern times) depends on `legacycustomsearches`. _But the `info.xml` doesn't declare the dependency._
* When you enable class-scanning, then it will try to read that class-file during installations. But the class cannot be read because `legacycustomsearches` is disabled.

I'll throw up a PR to see if it's easy to drop `generate:search`. (If it's hard to remove, then we can also fix it -- with a silver-lining of having an example of auto-updating the `<requires>` directive.)

totten:
civibot, test this please

PR #412, opened 2025-09-24
Add upgrade to add scan-classes mixin for extensions with Api4 classes

colemanw:
This used to be a soft-requirement but is now strongly encouraged.

- See https://github.com/civicrm/civicrm-core/pull/33371
- See https://github.com/totten/civix/pull/411

### DIFF
--- a/upgrades/25.09.0.up.php
+++ b/upgrades/25.09.0.up.php
@@ -1,0 +1,23 @@
+<?php
+
+/**
+ * Upgrade Api4 classes to use scan-classes mixin
+ */
+return function (\CRM\CivixBundle\Generator $gen) {
+
+  $checker = \Civix::checker();
+
+  if (is_dir($gen->baseDir->string('Civi/Api4')) && !$checker->hasMixin('/^scan-classes@/')) {
+    \Civix::io()->note([
+      'Extensions with Api4 classes should use the mixin `scan-classes`.',
+      'This will automatically add `scan-classes@1.0` since this extension has a Civi/Api4 directory.',
+      'See https://github.com/civicrm/civicrm-core/pull/33371',
+    ]);
+    if (!\Civix::io()->confirm('Continue with upgrade?')) {
+      throw new \RuntimeException('User stopped upgrade');
+    }
+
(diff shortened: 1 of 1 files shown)


colemanw:
civibot, test this please

totten:
My current hope is that https://github.com/civicrm/civicrm-core/pull/33680 will let us phase-out `LegacyEntityScanner` without any change in the downstream contract. And then we can tweak this upgrade to make it optional.

The cases where you wouldn't want to auto-enable are:

1. You have `*.php` files in `CRM/` or `Civi/` which are _not_ classes.
2. You have `*.php` class-files in `CRM/` or `Civi/` with a _soft-dependency on another project_. (Ex: For example, `cividiscount` and `oauth-client` provide base-classes. Your downstream extension may extend a base-classes _just-in-case_ -- but your extension is supposed to work without it.)

colemanw:
@totten given the improvement in https://github.com/civicrm/civicrm-core/pull/33680 I think this could be changed to

1. Check if the extension is implementing a hook that would take the place of `scan-classes`.
2. If so, offer to add the mixin very softly with caveats about the existing hook.
3. If not, offer and recommend adding the mixin.

totten:
@colemanw That sounds good to me.

FWIW, if you need to inspect or manipulate a function in the main PHP file, then [PrimitiveFunctionVisitor](https://github.com/totten/civix/blob/db7227c6e9ff2fe9d50354f2e761e09c1a1ba518/src/CRM/CivixBundle/Parse/PrimitiveFunctionVisitor.php#L5-L29) is pretty powerful. There's an example in `25.01.0.up.php`. The contours are like:

```php
$gen->updateModulePhp(function (\CRM\CivixBundle\Builder\Info $infoXml, string $body) {

  $hookFunc = $infoXml->getFile() . '_civicrm_scanClasses';
  return PrimitiveFunctionVisitor::visit($body, function (string &$func, string &$sig, string &$code) use ($hookFunc) {
    if ($func === $hookFunc) {
      // Inspect or manipulate
    }
  });

});
```

totten:
Replaced by #415.

PR #413, opened 2025-09-25
`generate:search` - Drop generator for deprecated subsystem

totten:
The `generate:search` command creates a skeletal subclass based on `CRM_Contact_Form_Search_Custom_Base`.

However, the entire subsystem was deprecated circa 2021 (~5.41), when it started moving into a core-extension (`legacycustomsearches`). The deprecation became more meaningful with the evolution of "Search Kit" -- there is now little reason for anyone to use `generate:search`.

Additionally, the generator is now a bit quirky -- it will create class-files, but you have to manually enable the dependency `legacycustomsearches` (for 5.41+).

Let's see if it's easy to remove...


colemanw:
@totten looks like it was straightforward to remove 😁

PR #427, opened 2025-12-17
Update civimix schema to @5.94.beta1

colemanw:
Adds change from https://github.com/civicrm/civicrm-core/pull/34286


colemanw:
@totten so should this be merged?

PR #434, opened 2026-02-12
[REGRESSION] Prevent using ts() on empty strings or NULL

colemanw:
Empty strings and null values should not be wrapped in `ts()`.
This looks to have regressed with 23a7e41537edde1771d9dd7cef670b42e900a732.


totten:
Looks good.

I coerced a use-case to provoke, and it worked as described.

PR #437, opened 2026-02-26
PHPUnit - Set CIVICRM_TEST constant when booting

colemanw:
See https://github.com/civicrm/civicrm-core/pull/34917


colemanw:
@totten on https://github.com/civicrm/civicrm-core/pull/34917 you said

> If it works across all the core-exts, then I think that validates the concept -- and it should be added to civix template.

So... merge this.

PR #440, opened 2026-06-01
Exclude settings from SearchDisplay exports

colemanw:
Settings should not be translated as of https://github.com/civicrm/civicrm-core/pull/33901

See https://github.com/civicrm/civicrm-core/pull/35782 for a similar patch to the Api4 Explorer.

- Introduced `excludeTs()` method to PhpData for skipping translation wrapping on specified keys and their children.
- Updated tests to verify exclusion behavior with nested and multiple keys.

### DIFF
--- a/src/CRM/CivixBundle/Builder/PhpData.php
+++ b/src/CRM/CivixBundle/Builder/PhpData.php
@@ -32,11 +32,13 @@ class PhpData implements Builder {
 
   /**
    * @var string[]
+   *   Ex: ['title', 'description'] means "Translate individual fields named `title` or `description`"
    */
   private $keysToTranslate;
 
   /**
    * @var string[]
+   *   Ex: ['extra'] means "Skip translation for subtrees named `extra`; e.g. skip `extra.title`".
    */
   private $keysToExcludeTs = [];
 
--- a/src/CRM/CivixBundle/Builder/PhpData.php
+++ b/src/CRM/CivixBundle/Builder/PhpData.php
@@ -4,10 +4,8 @@ namespace CRM\CivixBundle\Builder;
 use CRM\CivixBundle\Builder;
 use CRM\CivixBundle\Utils\Files;
 use CRM\CivixBundle\Utils\Path;
-use PhpArrayDocument\ArrayItemNode;
 use PhpArrayDocument\PhpArrayDocument;
 use PhpArrayDocument\Printer;
-use PhpArrayDocument\ScalarNode;
 use Symfony\Component\Console\Output\OutputInterface;
 
 /**
@@ -36,6 +34,11 @@ class PhpData implements Builder {
    * @var string[]
    */
   private $keysToTranslate;
+
+  /**
+   * @var string[]
+   */
+  private $keysToExcludeTs = [];
 
   /**
    * @var string
@@ -97,6 +100,16 @@ class PhpData implements Builder {
   }
 
   /**
+   * Specify fields (and their children) that will not be wrapped in E::ts()
+   *
+   * @param array $keysToExclude
+   * @return void
+   */
+  public function excludeTs(array $keysToExclude) {
+    $this->keysToExcludeTs = $keysToExclude;
+  }
+
+  /**
    * Adds `use Foo_ExtensionUtil as E;` to the top of the file
    *
    * @param string $extensionUtilClass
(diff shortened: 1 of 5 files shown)


colemanw:
@totten I don't think the test failures are caused by this PR.

colemanw:
@totten

totten:
> I don't think the test failures are caused by this PR.

Hmm... yeah, on a quick spot-check, the failures don't exactly make sense for the PR. OTOH, [contemporaneous nightly tests](https://test.civicrm.org/duderino/?filter=Civix-Periodic) are running clean. ([example](https://test.civicrm.org/duderino/batch/Civix-Periodic/20887518))  I guess it needs a closer look...

> Settings should not be translated as of https://github.com/civicrm/civicrm-core/pull/33901

Looking at this from downstream developer POV...  Does it seem right to say that  33901 represents an inversion of the best-practice for 6.16+?

* Before 6.16, you MUST use `E::ts()` in the SavedSearch `settings` if you want -any- i18n support?
* After 6.16, you MUST NOT use `E::ts()` in the SavedSearch `settings` if you want -good- i18n support?

Obviously, developers can be circumstantially indifferent re:i18n and re:compatibility. The question arises for developers who are diligent about both. And in this context, it sounds like...

| MGD Code | CiviCRM Version | CiviCRM Outcome <br/> (Single-lingual i18n) | CiviCRM Outcome <br/> (Multi-lingual i18n)
| -- | -- | -- | --
| 'text' => E::ts('Add Campaign')` | 6.15 |  🟢 Translates OK | 🟠  Erratic translation
| 'text' => E::ts('Add Campaign')` | 6.16  | 🟢 Translates OK | 🟠 Erratic translation
| 'text' => 'Add Campaign'` | 6.15  |  🔴 No translation | 🔴 No translation
| 'text' => 'Add Campaign'` | 6.16 |  🟢 Translates OK |   🟢 Translates OK

Is that the game?

colemanw:
@totten developers who are diligent will know when to add ts themselves. Also, there's no way to mark an extension as compatible with 6.15 but NOT compatible with 6.16 so... I vote we merge this PR as-is and not overthink it.

totten:
> developers who are diligent will know when to add ts themselves...

I really don't think it's that obvious.

> there's no way to mark an extension as compatible with 6.15 but NOT compatible with 6.16 so... I vote we merge this PR as-is and not overthink it.

It sounds like you've started thinking about a specific solution/approach/mitigation (re: details of publishing releases with compatibility flags) and started judging that. Maybe that rabbit-hole is good or bad, but right now I'm not in that rabbit-hole.

The place where I'm at: _The basic facts look unusual. Did we really invert the best-practice without a bridge? I'm not saying that's good/bad/right/wrong... it's probably good in this case... but it is unusual, and I want to be sure that it really happened._

colemanw:
@totten I'm glad you're not in that rabbit hole, but it really would have been great to raise these concerns on [the original PR](https://github.com/civicrm/civicrm-core/pull/33901) that has been open since last October. Better late than never I guess, but instead of having this conversation on a side channel we ought to be discussing on the original PR so the author can participate too.

src/CRM/CivixBundle/Builder/PhpData.php
```diff
@@ -151,16 +173,27 @@ public function save(&$ctx, OutputInterface $output) {
...
-      if (in_array($arrayItem->getKey(), $this->keysToTranslate ?: [], true) && $arrayItem->getValue() instanceof ScalarNode) {
+      // Check if this item or any of its ancestors should be excluded from ts
+      $shouldExcludeTs = FALSE;
+      $current = $arrayItem;
+      while ($current !== NULL) {
+        if (in_array($current->getKey(), $this->keysToExcludeFromTs, TRUE)) {
+          $shouldExcludeTs = TRUE;
+          break;
+        }
+        $current = $current->getParent();
```
totten: @colemanw  -All of the tests are failing because of this call to `getParent()`. I don't see any implementation of `getParent()` in `php-array-doc.git` (*either `master` or in pending PRs*).

(*UPDATE: Example error*)
```
Write info.xml
Write Civi/Api4/MyEntityFour.php
Write CRM/Civixsnapshot/BAO/MyEntityFour.php
Write schema/MyEntityFour.entityType.php
Error: Call to undefined method CivixPhar\PhpArrayDocument\ArrayItemNode::getParent() in CRM\CivixBundle\Builder\PhpData->save() (line 170 of phar:///home/homer/buildkit/build/build-1/src/civix/bin/civix.phar/src/CRM/CivixBundle/Builder/PhpData.php).
```

Could there be an uncommitted local change?

colemanw: Whoops. That was it.

colemanw:
test this please

colemanw:
Thanks @totten all fixed.

totten:
> Thanks @totten all fixed.

Woot

> I'm glad you're not in that rabbit hole, but it really would have been great to raise these concerns on https://github.com/civicrm/civicrm-core/pull/33901 that has been open since last October. Better late than never I guess, but instead of having this conversation on a side channel we ought to be discussing on the original PR so the author can participate too.

Perhaps -- but just to clarify, I'm not trying to litigate an alternative to 33901. Sometimes a change is necessary+acceptable. I don't really see anything that provides interop while leaving the system as clean. (Part of the benefit of 33901 is that the final arrays are actually a bit cleaner...) And in the worst-case, bad choice here doesn't leave you with a broken system -- it just means some of the strings won't be translated.

From a civix POV, I just want to communicate the migration effectively.

Added upgrade step to notify developer and (conditionally) drop all the old `E::ts()` clauses.

colemanw:
@totten great, all sounds good. Thanks for adding the upgrade steps.

Can you please cut a new release?

PR #441, opened 2026-06-04
dev/core#6550 Update type hints to support upgrading

eileenmcnaughton:
The type hint is required when upgrading symfony and it is Ok to have it with the current version


colemanw:
@totten let's merge this.

totten:
Yep, lgtm

PR #446, opened 2026-08-28
Only check for 5.74 at runtime if min compatibility is lower

colemanw:
## Overview
In `module.civix.php.php`, only emit the runtime `version_compare(CRM_Utils_System::version(), '5.74.beta', '>=')` check for DAO base class alias if `$_compatibility` is less than `'5.74.beta'`.

## Technical Details
When an extension's `<compatibility>` target is already `>= 5.74.beta`, we know CiviCRM will always be at least `5.74.beta`, so `class_alias('CRM_Core_DAO_Base', ...)` can be called directly without wrapping it in an unnecessary runtime version check.


totten:
This seems reasonable, off the top of my head.

Fired some adhoc tests (above and below 5.74 -- https://test.civicrm.org/duderino/batch/Adhoc/20890971), and they're also passing.

